PR #2, opened 2021-04-26
Add draft Excel-to-sbol3 utility

jakebeal:


cjmyers:
@jakebeal do you want this merged or continuing to add to it?

jakebeal:
I wanted to check with @bbartley to make sure this is matching his pattern for the utilities.

cjmyers:
Ok, as a protocol, perhaps add reviewers to those you want to review before it is merged.  This way the editors and I will know when you want things merged.  Of course, on repos like these you likely can just merge yourself, if you like.

sbol_utilities/__init__.py
```diff
@@ -0,0 +1 @@
+from .excel_to_sbol import excel_to_sbol
```
bbartley, line 1: This line is not necessary at this point.  Even without this line, you can still import it as `from sbol_utilities.excel_to_sbol import excel_to_sbol`.  The reason that I lean toward this latter pattern (although it appears a little redundant) is because I anticipate that each utility will eventually evolve to have multiple methods, and it will be nice to have these methods scoped to a particular utility namespace rather than cluttering the global namespace.

bbartley (commented):
I've never used `unittest` and `pytest` frameworks together (usually just one or the other).  Apparently, however, it is possible to use both together, according to the internets, which is new to me.  Since the tests can now be invoked by a variety of different commands, can you please put an example of the preferred invocation in the README?

sbol_utilities/__init__.py
jakebeal: Rather than the `from ... import` pattern, I've been trying to keep my global namespace clean by doing `import package` and then invoking with `package.function`. With that pattern, I was having to do `sbol_utilities.excel_to_sbol.excel_to_sbol`, which feels awkward.

jakebeal:
I've been running via pytest. My poor python knowledge had actually made me think I _needed_ both. I definitely need pytest in order to check for exceptions. Can you recommend how to adjust to need only pytest?

sbol_utilities/__init__.py
bbartley: The situation I'm trying to avoid (and what I mean by cluttering the namespace) is this:
```
sbol_utilities.utilityA_method1
sbol_utilities.utilityA_method2
sbol_utilities.utilityB_method1
sbol_utilities.utilityB_method2
sbol_utilities.utilityC_method1
sbol_utilities.utilityC_method2
```
In this case, if I import `sbol_utilities` I get all of the methods jumbled together in the package namespace.  Whereas, what I'm advocating is a subpackage organization like this:
```
sbol_utilities.utilityA.method1
sbol_utilities.utilityA.method2
sbol_utilities.utilityB.method1
sbol_utilities.utilityB.method2
sbol_utilities.utilityC.method1
sbol_utilities.utilityC.method2
```
Now, if I `import sbol_utilities.utilityA`, I only import those methods associated with utilityA.

A less verbose solution for your example above is to import the subpackage as follows:
```
import sbol_utilities.excel_to_sbol as excel_to_sbol
excel_to_sbol.excel_to_sbol(arg0, arg1, ...)
```
While this may still seem a little redundant and silly at the moment, it will appear less so as the utilities expand to encompass more methods, e.g.:
```
excel_to_sbol.sbol_to_excel
excel_to_sbol.validate_spreadsheet
excel_to_sbol.merge_spreadsheets
```

bbartley:
I think all you have to do is remove the `TestExcelToSBOL` and convert the methods back to regular functions (remove the `self` argument).  I don't think you need the `if __name__ == '__main__' ` clause either

jakebeal:
I've made all of the requested changes.  I also figured out how to make excel-to-sbol a command-line utility without creating a separate file, which we'll likely want to do for graph-sbol as well.

bbartley (approved):

PR #4, opened 2021-05-04
Expand cds

jakebeal:
Utility to expand combinatorial derivations


sbol_utilities/helper_functions.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+#########################
+# Kludge materials that should be removed after certain issues are resolved
+
+
+# Workaround for pySBOL3 issue #231: should be applied to every iteration on a collection of SBOL objects
+def id_sort(i: iter):
+    sortable = list(i)
+    sortable.sort(key=lambda x: x.identity if isinstance(x, sbol3.Identified) else x.lookup().identity)
```
bbartley, line 21: The `else` clause seems kind of circuitous.  Wouldn't the following accomplish the same thing?
```
sortable.sort(key=lambda x: x.identity if isinstance(x, sbol3.Identified) else x)
```

test/test_expand_combinatorial_derivations.py
```diff
@@ -0,0 +1,62 @@
...
+    #
+    # assert not doc.validate().errors and not doc.validate().warnings
+    # assert len(doc.find('BasicParts').members) == 43
+    # assert len(doc.find('CompositeParts').members) == 8
+    # assert len(doc.find('LinearDNAProducts').members) == 2
+    # assert len(doc.find('FinalProducts').members) == 2
+    #
+    # temp_name = tempfile.gettempdir() + '/' + next(tempfile._get_candidate_names())
+    # doc.write(temp_name, sbol3.SORTED_NTRIPLES)
+    # assert filecmp.cmp(temp_name, TESTFILE_DIR + '/constraints_library.nt')
```
bbartley, line 54: Is this intended to work with a future upgrade?  If so I would document it as such. Otherwise, if it is completely non-functional, I suggest removing it.

sbol_utilities/helper_functions.py
```diff
@@ -0,0 +1,71 @@
...
+    # should be more thorough, but kludging to just look at constraints
+    for ct in component.constraints:
+        if ct.subject == old.identity: ct.subject = new.identity
+        if ct.object == old.identity: ct.object = new.identity
+
+
+########################
+# Used in testing; TODO: figure out how to host it in a test helpers file rather than the package
+# check if two files are identical; if not, report their diff
+def assert_files_identical(file1, file2):
```
bbartley, line 66: I think Tom has a similar method in the pysbol tests.  We might want to pull it out and make it a method on `Document`?   It seems to be a not unusual operation that users will want to perform.

sbol_utilities/helper_functions.py
jakebeal: I don't think it will, because in that case x is not a string but a ReferencedURI

test/test_expand_combinatorial_derivations.py
jakebeal: Yes, this matches with a TODO in the file.

bbartley (changes requested):
See line 21 in helper_functions.py and line 41 in test_expand_combinatorial_derivations.py  for a couple of nits that should be addressed.

A couple of stylistic notes for future reference that you don't necessarily need to address with changes right now.
- Documentation for a function or method usually comes immediately _after_ the declaration and uses docstrings `"""My function serves this purpose"""`
- Might want to use type hints on functions return values as well as arguments, e.g.: `def cd_assigment_to_display_id(cd: sbol3.CombinatorialDerivation, assignment: tuple) -> bool:`
- A lot of terse logic can be packed into list comprehensions, but sometimes at the expense of readability

sbol_utilities/helper_functions.py
jakebeal: I don't think it makes sense on `Document` since it's specific to a serialization. But if there's a good place to stick reusable test infrastructure that would be great.

sbol_utilities/helper_functions.py
bbartley: ReferencedURI should return a string

jakebeal: Just tested it, and it looks like you are correct.

bbartley (approved):

PR #45, opened 2021-10-12
Add converters

jakebeal:
Add SBOL conversion package, imported from the implementation in iGEM distribution
Includes both a generic command-line utility and pairwise command-line utilities

### DIFF
--- a/sbol_utilities/conversion.py
+++ b/sbol_utilities/conversion.py
@@ -79,7 +79,14 @@ def convert2to3(sbol2_doc: Union[str, sbol2.Document], namespaces=None) -> sbol3
         namespaces = []
     if isinstance(sbol2_doc, sbol2.Document):
         sbol2_path = tempfile.mkstemp(suffix='.xml')[1]
-        sbol2_doc.write(sbol2_path)
+        # Turn off automatic validation on write to avoid requiring transmitting data to the validator webservice
+        # If you want validation, you should validate the document explicitly beforehand
+        was_using_validator = sbol2.Config.getOption(sbol2.ConfigOptions.VALIDATE)
+        try:
+            sbol2.Config.setOption(sbol2.ConfigOptions.VALIDATE, False)
+            sbol2_doc.write(sbol2_path)
+        finally:
+            sbol2.Config.setOption(sbol2.ConfigOptions.VALIDATE, was_using_validator)
     else:
         sbol2_path = sbol2_doc
 
@@ -185,7 +192,8 @@ def convert3to2(doc3: sbol3.Document) -> sbol2.Document:
         for sa in c.sequenceAnnotations:
             for loc in sa.locations:
                 loc.sequence = None  # remove optional sequences, per https://github.com/SynBioDex/libSBOLj/issues/621
-    doc2.validate()
+    # We explicitly do NOT validate here in order to avoid requiring transmitting data to the validator webservice
+    # If you want validation, you should run it on the document that is returned
     return doc2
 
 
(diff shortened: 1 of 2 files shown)


.github/workflows/python-app.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       matrix:
         # Default builds are on Ubuntu
         os: [ubuntu-latest]
-        python-version: [3.7, 3.8, 3.9, pypy-3.7]
+        python-version: [3.7, 3.8, 3.9]
```
tcmitchell, line 20: I recently learned that these should be strings. When you add Python 3.10 to the list it will have to be a string.

```suggestion
        python-version: ['3.7', '3.8', '3.9']
```

README.md
```diff
@@ -34,6 +34,15 @@ graph-sbol -i my_file.ttl
 
 The `excel-to-sbol` utility reads an Excel file specifying a library of basic and composite parts, formatted following `sbol_library_template.xlsx`.
 
+### Convert beween SBOL3 and other genetic design formats
+
+The `sbol-converter` utility converts between any of the SBOL3, SBOL2, GenBank, and FASTA formats.
+
+Additional "macro" utilities convert specifically between SBOL3 and one of the other formats: 
+- `sbol2fasta` and `fasta2sbol` convert to and from FASTA
```
tcmitchell, line 42: It looks like these convert between SBOL 3 and FASTA, but the name `sbol2fasta` sort of looks like it might do SBOL2 conversion. It might be useful to explicitly state the conversion is between SBOL 3 and FASTA.

README.md
```diff
@@ -34,6 +34,15 @@ graph-sbol -i my_file.ttl
...
+- `sbol2genbank` and `genbank2sbol` convert to and from GenBank
```
tcmitchell, line 43: It looks like these convert between SBOL 3 and FASTA, but the name `sbol2genbank` sort of looks like it might do SBOL2 conversion. It might be useful to explicitly state the conversion is between SBOL 3 and FASTA.

.github/workflows/python-app.yml
tcmitchell, line 20: I'm curious, why was pypy-3.7 dropped? I don't think that's a bad thing, just curious.

tcmitchell, line 20: Python 3.10 has been released. Consider adding it to the list. 3.7 -> 3.10 are all currently supported. 3.6 is nearly end of life, so there's no need to support it.

tcmitchell (changes requested):
It's exciting to have these utilities more broadly available!

.github/workflows/python-app.yml
jakebeal: I dropped this because I started running into build problems with its dependencies, which I took to be something related to use of the sbol2 package. Now that I've got other issues resolved, I'll try it in one more before the merge and see if it can go forward or not.

README.md
jakebeal: Done.

README.md
jakebeal: Done.

sbol_utilities/conversion.py
```diff
@@ -0,0 +1,465 @@
...
+    """Convert a GenBank document on disk into an SBOL3 document
+    Specifically, the GenBank document is first imported to SBOL2, then converted from SBOL2 to SBOL3
+
+    :param path: path to read GenBank file from
+    :param namespace: URIs of Components will be set to {namespace}/{genbank_id}
+    :return: SBOL3 document containing converted materials
+    """
+    doc2 = sbol2.Document()
+    sbol2.setHomespace(namespace)
+    doc2.importFromFormat(path)
```
tcmitchell, line 248: Eek! This sends the input file over the internet.

You may want to add an argument to `convert_from_genbank` that forces the user to say `online_conversion=True`, and it defaults to `False`, and if it is False, raise an error.

In the future, perhaps we can use BioPython to do local conversion. I don't know how easy or hard that would be.

tcmitchell (changes requested):

PR #57, opened 2021-11-07
default components for pySBOL3 high level scripting

Gonza10V:
name still not decided
some functions are still drafts
Im open to add more components
need to discuss default parameters


sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
+from typing import List, Tuple
```
jakebeal, line 1: Could this module be combined with the existing `sbol_utilities.component` module?

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+
+import sbol3
+import tyto
+
+def promoter(name: str, description: str, sequence: str)-> Tuple[sbol3.Component, sbol3.Sequence]:
+    """
+    Creates a promoter and a sequence Components with a given name, description, and sequence.
```
jakebeal, line 8: In this an all of the other routines, the docstring should include input and output argument information as well. See the examples in `sbol_utilities.component`

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
+from typing import List, Tuple
+
+import sbol3
+import tyto
+
+def promoter(name: str, description: str, sequence: str)-> Tuple[sbol3.Component, sbol3.Sequence]:
```
jakebeal, line 6: - Is this really name, or should it be display_id?
- Can you allow all of the additional arguments to be potentially passed through via kwargs?
- I am guessing that the sequence is assumed to be IUPAC format; that should be documented.
- Should description really be a required argument, and not a kwarg?

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    Creates a promoter and a sequence Components with a given name, description, and sequence.
+    """
+    promoter_seq = sbol3.Sequence(f'{name}_seq')
+    promoter_seq.elements= sequence
+    promoter_seq.encoding = 'https://identifiers.org/edam:format_1207'
```
jakebeal, line 12: Use `sbol3.IUPAC_DNA_ENCODING`

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+    promoter.description = description
+    promoter.roles. append(sbol3.SO_PROMOTER)
+    promoter.sequences.append(promoter_seq)
+    return tuple([promoter, promoter_seq])
+
+def rbs(name: str, description: str, sequence: str)-> Tuple[sbol3.Component, sbol3.Sequence]:
+    """
+    Creates a Ribosome Binding Site (RBS) and a sequence Components with a given name, description, and sequence.
+    """
+    rbs_seq = sbol3.Sequence(f'{name}_seq')
```
jakebeal, line 25: There's a lot of repeated structure in these functions. The common structure should be factored out into a shared function (`dna_component_with_sequence`?), such that each of these functions is a single line of code plus documentation. That will also make it easier to maintain and expand the collections.

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+    terminator_seq.encoding = 'https://identifiers.org/edam:format_1207'
+
+    terminator = sbol3.Component(name, sbol3.SBO_DNA)
+    terminator.name = name
+    terminator.description = description
+    terminator.roles. append(sbol3.SO_TERMINATOR)
+    terminator.sequences.append(terminator_seq)
+    return tuple([terminator, terminator_seq])
+
+def deg_tag(name: str, description: str, sequence: str)-> Tuple[sbol3.Component, sbol3.Sequence]:
```
jakebeal, line 66: It's odd to have this one be abbreviated while all the others match their SO names.

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+        engineered_region.features.append(x)
+    #fix order
+    if len(engineered_region.features) > 1:
+            for i in range(len(engineered_region.features)-1):
+                engineered_region.constraints = [sbol3.Constraint(sbol3.SBOL_PRECEDES, engineered_region.features[i], engineered_region.features[i+1])]
+    else: pass
+    #engineered_region.sequences.append(engineered_region_seq)
+    return engineered_region # , engineered_region_seq
+
+def media(name: str, description: str)-> sbol3.Component:
```
jakebeal, line 120: How about having the recipe for the media optionally passed as a `Dict(Component,Measure)`?

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    Creates a strain Component with a given name and description.
+    """
+    strain = sbol3.Component(name, sbol3.SBO_FUNCTIONAL_ENTITY)
+    strain.name = name
+    strain.description = description
+    strain.roles.append(tyto.NCIT.Strain)
+    return strain
+
+def simple_chemical(name: str, description: str, chebi: str )-> sbol3.Component:
```
jakebeal, line 140: ChEBI isn't the only way to make a simple chemical. There's also PubChem, SMILES, etc.
I'm also wondering if this would do better as a way to add a LocalSubComponent to a system Component, rather than stand-along definitions when we can avoid it.

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    Creates a media Component with a given name and description.
+    """
+    media = sbol3.Component(name, sbol3.SBO_FUNCTIONAL_ENTITY)
+    media.name = name
+    media.description = description
+    media.roles.append(tyto.NCIT.Media)
+    return media
+
+def strain(name: str, description: str)-> sbol3.Component:
```
jakebeal, line 130: Don't you want to add any other information about a strain?

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+    plasmid_vector_seq.encoding = 'https://identifiers.org/edam:format_1207'
+
+    plasmid_vector = sbol3.Component(name, sbol3.SBO_DNA)
+    plasmid_vector.name = name
+    plasmid_vector.description = description
+    plasmid_vector.roles. append(tyto.SO.plasmid_vector)
+    plasmid_vector.sequences.append(plasmid_vector_seq)
+    return tuple([plasmid_vector, plasmid_vector_seq])
+
+def engineered_region(name: str, description: str, sub_components: List[sbol3.SubComponent])-> sbol3.Component:
```
jakebeal, line 96: What defines the set of the things that have been chosen for short-cut functions vs. things that do not have them?

jakebeal (changes requested):

sbol_utilities/default_components.py
Gonza10V: Thanks I could not found that, I will replace the URI

sbol_utilities/default_components.py
Gonza10V: Yes they have the same pattern, the only difference is the role, that can be passed. How can we pass the argument ? It could be the URI 
```
ptet, ptet_seq = dna_component_with_sequence('sbol3.SO_PROMOTER', 'L3S2P21_UPA20_pTetR_RiboJ', 'BASIC - Potvin pTet (based on B-P27-pTetR-F1)','GTCCCTCG')
```
or as text that has a dictionary from  strings to URIs like {'promoter' : sbol3.SO_PROMOTER}. 
```
ptet, ptet_seq = dna_component_with_sequence('promoter', 'L3S2P21_UPA20_pTetR_RiboJ', 'BASIC - Potvin pTet (based on B-P27-pTetR-F1)','GTCCCTCG')
```
How does that affect the code readability. 
```
ptet, ptet_seq = promoter('L3S2P21_UPA20_pTetR_RiboJ', 'BASIC - Potvin pTet (based on B-P27-pTetR-F1)','GTCCCTCG')
```
I think the actual implementation is the most readable of the options but I never thought on maintainment nor its relation to collections. Can you explain me better those aspects please?

sbol_utilities/default_components.py
Gonza10V: Agree if they have a name it should be the SO names I will change this function name for protein_stability_element.

sbol_utilities/default_components.py
Gonza10V: I like the idea of having the recipe of the media. How can we store that info into SBOL3? My first idea is to create Components  of the recipe items and append the Measure to that Component measures , then make SubComponents from the recipe items and append them as features on the media Component. What do you think?

sbol_utilities/default_components.py
Gonza10V: If I understand it correct, the best way to implement this is through the SEP054? If that is the case I agree that a conversion package is the best solution. How long could it take to be implemented? How can we use a function like this on the meantime?

sbol_utilities/default_components.py
Gonza10V: hmmm I dont know what could be the best. Maybe biocyc, but I dont know if produces stable URIs, that is an option. In Newcastle are developing a kind of GitHub for strains and engineered organisms, I will look for more information, I dont remember the name right now.

sbol_utilities/default_components.py
Gonza10V: That is a very good question, I do it arbitrarily as they appear on my workflow but we could set a rational line, establish the things to be included and develop the package. My naive approximation is to start by all the things that we use in our workflows.

sbol_utilities/default_components.py
jakebeal: I think that the current interface can stay the same. I am suggesting that the function `promoter` be implemented by a call to `dna_component_with_sequence`.  That way, `promoter`, `rbs`, `terminator`, etc do not contain any repeated code.

sbol_utilities/default_components.py
jakebeal: Section 7.9.1 of the specification already provides this as a recommended practice.

I would suggest that in some cases it would make sense to use SubComponent and in others to use ExternalReference. For some URIs, it would be obvious which to use (e.g., ChEBI is ExternalReference, a Component in the document is SubComponent). For other URIs, it would not be obvious, however, and might need to be provided by the user in some manner, e.g., a wrapping function.

sbol_utilities/default_components.py
jakebeal: Above, I meant to say ExternallyDefined instead of LocalSubComponent.  Those don't need SEP_054 -- the whole point of ExternallyDefined is to allow us to do this with just an ontology term (which can be accessed via tyto), rather than needing to create a TopLevel.

sbol_utilities/default_components.py
jakebeal: Take a look at Section 7.10 in the specification, which has the current best practice recommendations (which were developed in part by Newcastle).

sbol_utilities/default_components.py
jakebeal: I might suggest that everything that gets an explicit call-out in the specification is a good target, e.g., Table 4, the best practices sections I've referenced above, and other similar tables.

sbol_utilities/default_components.py
Gonza10V: Ah, now I understand, to make the repeated pattern in a function that then gets used by all the others. I will implement that.

sbol_utilities/default_components.py
Gonza10V: Yes, I will add the functions to `sbol_utilities.component` module

sbol_utilities/default_components.py
Gonza10V: I will improve the docstrings following the `sbol_utilities.component`

sbol_utilities/default_components.py
Gonza10V: > Is this really name, or should it be display_id?

In the functions name and display_id are the same. Does that produce a problem? or if I state that on the docstring should be fine? Another solution is to use `display_id = f'{name}_id'`

> Can you allow all of the additional arguments to be potentially passed through via kwargs?

Sure, which arguments do you want to add?

> I am guessing that the sequence is assumed to be IUPAC format; that should be documented.

Should it be on the docstring?

> Should description really be a required argument, and not a kwarg?

Good point, the only required arguments should be the ones required by pySBOL3

jakebeal: In terms of parameters: we _have_ to have an identity. That might be provided as a display_id or it might be provided as a full URI. Your functions don't need to distinguish whether they've gotten a full identity or a display_id because in either case it's valid for the constructors and the extension with `_seq`.

For other arguments, you don't need to explicitly provide them: you can use the python kwargs functionality to accept an unbounded set of keyword arguments and to pass them through to the Component constructor. The only arguments that need to be explicit are the ones that get manipulated in these functions.

Gonza10V: What if instead of name I use identity and then explain in the docstring that function also use that for name? Name could not be included on the function since it is not a required parameter or the function can use identity as name if name is not provided. I would go for the last one.
Do you have a source or an example of how to implement kwargs, I've never used them before.

jakebeal: I think it would work to just use identity and allow name to be passed in the kwargs parameter if desired.
The defaulting of an unset name to display_id is already separately proposed as a practice in https://github.com/SynBioDex/pySBOL3/issues/229

On kwargs, here's a tutorial: https://www.geeksforgeeks.org/args-kwargs-python/
You can ask the internet / StackExchange for additional help

sbol_utilities/default_components.py
Gonza10V: OK, here is a draft of that implementation, with using `dna_component_with_sequence` 

```
    def dna_component_with_sequence(identity: str, sequence: str, *, description: str)-> Tuple[sbol3.Component, sbol3.Sequence]:
    """
    Creates a Component and its Sequence.

    :param name: The name and identity of the Component. The identity of Sequence is also name with the suffix '_seq'.
    :param sequence: The DNA sequence of the Component encoded in IUPAC.
    :param description: The description of the Component.
    :return: A tuple of Component and Sequence.
    """
    comp_seq = sbol3.Sequence(f'{identity}_seq')
    comp_seq.elements= sequence
    comp_seq.encoding = 'sbol3.IUPAC_DNA_ENCODING'

    dna_comp = sbol3.Component(identity, sbol3.SBO_DNA)
    dna_comp.name = identity
    dna_comp.description = description
    dna_comp.sequences.append(comp_seq)
    return tuple([dna_comp, comp_seq])
```

Then to build a promoter the function should be like this:
```
    def promoter(identity: str, sequence: str, *, description: str)-> Tuple[sbol3.Component, sbol3.Sequence]:
    """
    Creates a Promoter Component and its Sequence.

    :param name: The name and identity of the Component. The identity of Sequence is also name with the suffix '_seq'.
    :param sequence: The DNA sequence of the Component encoded in IUPAC.
    :param description: The description of the Component.
    :return: A tuple of Component and Sequence.
    """
    promoter, promoter_seq = dna_component_with_sequence(identity, sequence, description=description)
    promoter.roles.append(sbol3.SO_PROMOTER)
    return tuple([promoter, promoter_seq])
```

sbol_utilities/default_components.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    promoter_seq = sbol3.Sequence(f'{name}_seq')
+    promoter_seq.elements= sequence
+    promoter_seq.encoding = 'https://identifiers.org/edam:format_1207'
+
+    promoter = sbol3.Component(name, sbol3.SBO_DNA)
+    promoter.name = name
+    promoter.description = description
+    promoter.roles. append(sbol3.SO_PROMOTER)
+    promoter.sequences.append(promoter_seq)
```
tcmitchell, line 18: These could be done in a single constructor call. For example (this might be a little off syntactically so you may have to adjust it, I'm typing from memory):
```python
promoter = sbol3.Component(name, sbol3.SBO_DNA,
                           name=name,
                           description=description,
                           roles=[sbol3.SO_PROMOTER],
                           sequences=[promoter_seq])
```

tcmitchell:
Somewhere, maybe in an email thread, you mentioned adding things to a document as needing to be separate calls. This has changed recently in pySBOL3. `Document.add` takes a single top level or a sequence of TopLevel objects. So you can do a single add to get all of the produced items into the document.

I don't see an `add` in the pull request, so maybe that would go in example code or documentation or something like that.

tcmitchell:
@Gonza10V I really like these functions. Thank you for taking the time to write them and contribute them. These are among the features needed to make pySBOL3 useful. Thanks!

tcmitchell (commented):
My only comment is about the possible use of the keyword arguments to the constructors.

sbol_utilities/default_components.py
jakebeal: Yes, that's along the lines I'm thinking, plus the keyword arguments.

Gonza10V: Ok here is the implementation using kwargs and @tcmitchell constructor suggestion:
```
def dna_component_with_sequence(identity: str, sequence: str, **kwargs)-> Tuple[sbol3.Component, sbol3.Sequence]:
    """
    Creates a Component and its Sequence.

    :param identity: The identity of the Component. The identity of Sequence is also identity with the suffix '_seq'.
    :param sequence: The DNA sequence of the Component encoded in IUPAC.
    :param **kwargs: Any other Component attribute.
    :return: A tuple of Component and Sequence.
    """
    comp_seq = sbol3.Sequence(f'{identity}_seq',
                                elements=sequence,
                                encoding='sbol3.IUPAC_DNA_ENCODING')

    dna_comp = sbol3.Component(identity, sbol3.SBO_DNA,
                                sequences =[comp_seq],
                                **kwargs)
    return tuple([dna_comp, comp_seq])

def promoter(identity: str, sequence: str, **kwargs)-> Tuple[sbol3.Component, sbol3.Sequence]:
    """
    Creates a Promoter Component and its Sequence.

    :param identity: The identity of the Component. The identity of Sequence is also identity with the suffix '_seq'.
    :param sequence: The DNA sequence of the Component encoded in IUPAC.
    :param **kwargs: Any other Component attribute.
    :return: A tuple of Component and Sequence.
    """
    promoter, promoter_seq = dna_component_with_sequence(identity, sequence, **kwargs)
    promoter.roles.append(sbol3.SO_PROMOTER)
    return tuple([promoter, promoter_seq])
```

how should I indicate to the user how to use the kwarg? I used :param **kwargs: Any other Component attribute.

jakebeal: That's looking pretty good to me.
I don't know the most pythonic way of documenting kwargs; ask the internet?

sbol_utilities/default_components.py
Gonza10V: > Somewhere, maybe in an email thread, you mentioned adding things to a document as needing to be separate calls. This has changed recently in pySBOL3. `Document.add` takes a single top level or a sequence of TopLevel objects. So you can do a single add to get all of the produced items into the document.
> 
> I don't see an `add` in the pull request, so maybe that would go in example code or documentation or something like that.

I saw the implementation on the pySBOL3 repo to add all, but apparently it was not implemented when I imported the package. I clone and use the pySBOL repo and that have the feature. Maybe it is a version problem.

sbol_utilities/default_components.py
Gonza10V: I have a proposal implementation now. The recipe would be a dictionary where the key is a Component and the value is a list of a number and a measure unit URI like this:
```
peptone = sbol3.Component('Bacto_Peptone', 
            tyto.SBO.functional_entity,
            name = 'Bacto_Peptone',
            derived_from = ['https://www.thermofisher.com/order/catalog/product/211820'])

nacl = sbol3.Component('NaCl', 
            tyto.SBO.functional_entity,
            name = 'NaCl',
            derived_from = ['https://www.sigmaaldrich.com/AU/en/product/sigald/s9888'])

yeast_extract = sbol3.Component('Yeast_Extract', 
            tyto.SBO.functional_entity,
            name = 'Yeast_Extract',
            derived_from = ['https://www.thermofisher.com/order/catalog/product/212720'])

recipe = {
    peptone:[10,tyto.OM.gram],
    nacl:[5,tyto.OM.gram],
    yeast_extract:[5,tyto.OM.gram]}
```

I could develop functions to make reactives or fungibles specifying identity and a URL that may need some recommendations or best practices to be more standard.

Then the function that creates media looks like this:
```
def media(identity: str, recipe = None, **kwargs)-> sbol3.Component:
    """
    Creates a media Component of type functional entity.

    :param identity: The identity of the Component.
    :param **kwargs: Keyword arguments of any other Component attribute.
    :return: A Component object.
    """
    media =  sbol3.Component(identity, sbol3.SBO_FUNCTIONAL_ENTITY, **kwargs)
    media.roles.append(tyto.NCIT.Media)
    if recipe:
        for key, value in recipe:
            if isinstance(key, sbol3.Component):
                key = sbol3.SubComponent(key)
            key.measures.append(sbol3.Measure(value[0], value[1]))
            media.features.append(key)
    return media
```

sbol_utilities/default_components.py
Gonza10V: It could be like this:
```
def ed_simple_chemical(definition: str, **kwargs)-> sbol3.Component:
    """
    Creates an ExternallyDefined Simple Chemical Component.

    :param definition: The URI that links to a canonical definitino external to SBOL, recommended ChEBI.
    :param **kwargs: Keyword arguments of any other ExternallyDefined attribute.
    :return: A Component object.
    """
    ed_simple_chemical = sbol3.ExternallyDefined([sbol3.SBO_SIMPLE_CHEMICAL], definition, **kwargs)
    return ed_simple_chemical
```

jakebeal: This looks reasonable. Would you consider providing the component that is to be its context to add directly into, like some of the existing functions do for things like interactions?

sbol_utilities/default_components.py
jakebeal: This looks like a fine first pass, though I think you might consider externally defined features as well.

sbol_utilities/default_components.py
Gonza10V: The SBOL Specification 3.0.1 section 7.10 RECOMMEND to the type “cell” from the Gene Ontology (GO:0005623). Looked for this term and it is [obsolete ](https://www.ebi.ac.uk/QuickGO/term/GO:0005623). I've raised an issue on SBOL specification to address it (https://github.com/SynBioDex/SBOL-specification/issues/474). The constructor now provides strain and the kwargs can handle any other information.

This is how looks the function:
```
def strain(identity: str, **kwargs)-> sbol3.Component:
    """
    Creates a strain Component of type functional entity.

    :param identity: The identity of the Component.
    :param **kwargs: Keyword arguments of any other Component attribute.
    :return: A Component object.
    """
    strain =  sbol3.Component(identity, sbol3.SBO_FUNCTIONAL_ENTITY, **kwargs)
    strain.roles.append(tyto.NCIT.Strain)
    return strain
```

sbol_utilities/default_components.py
Gonza10V: I've added all the example on table 4 but the effector since it could be externally defined as well, more components can be added later.

sbol_utilities/default_components.py
Gonza10V: This function does accept ExternallyDefined features, it just converts Components into SubComponents and if it is ExternallyDefined it just add the measure and append it. I just tested it with an ExternallyDefined object in the recipe. Was that the consideration?

sbol_utilities/default_components.py
Gonza10V: Sorry I didn't understand that, could you provide an example?

sbol_utilities/default_components.py
Gonza10V: Thanks @tcmitchell this helped me to implement kwargs.

sbol_utilities/default_components.py
jakebeal: Ah, I didn't read carefully enough.  Definitely add an example illustrating this in the regression tests to go with this pull.

sbol_utilities/default_components.py
jakebeal: In component, check out routines like add_feature, constitutive, and add_interaction for somewhat similar cases.

sbol_utilities/default_components.py
Gonza10V: Sure I can add an example, what and where is the regression tests?  Should I do something like this https://github.com/SynBioDex/SBOL-utilities/blob/develop/test/test_component.py ?

jakebeal: Correct. Every significant operation in the utilities should be tested via test functions, which are run automatically in https://github.com/SynBioDex/SBOL-utilities/actions Since you are adding to the component package, adding tests to test_component.py would be good.

Gonza10V: I implemented some test like the ones that I run on the terminal. I make components using the new functions and add them to the document, then I check if the document is valid or the length of the report == 0.

Gonza10V: @jakebeal Now the test is working properly, let me know if you have feedback, otherwise we can go ahead with the merge.

test/test_component.py
```diff
@@ -47,6 +47,15 @@ def test_system_building(self):
         comparison_file = os.path.join(test_dir, 'test_files', 'component_construction.nt')
         assert filecmp.cmp(tmp_out, comparison_file), f'Converted file {tmp_out} is not identical'
 
+        hlc_doc = sbol3.Document()
```
jakebeal, line 50: Three changes to request:
1) These should be in their own test function (e.g. `def test_high_level_constructors(self)`, not in `test_system_building`
2) The test should check that the _contents_ of the constructed components, not just that they validate. You can do this by writing a copy of the test document, checking it by hand for correctness, then comparing the results of your builder calls to make sure the document you produce is the same as the hand-checked test document. This is what is done with `test_system_building`
3) The test should exercise all of the new functions, not just a subset.

The first change will make it easier to debug failures.
The second and third changes are important to make sure that your functions are producing the desired result.

jakebeal (changes requested):

test/test_component.py
Gonza10V: @jakebeal to Implement the proposed test I though in something like this for every constructor:

```
dna_identity = 'dna_component_with_sequence'
dna_sequence = 'ttt'
test_description = 'test'
hl_dna_comp, hl_dna_seq = dna_component_with_sequence(dna_identity, dna_sequence, description=test_description)
dna_seq = sbol3.Sequence(f'{dna_identity}_seq',
                        elements=dna_sequence,
                        encoding=sbol3.IUPAC_DNA_ENCODING)

dna_comp = sbol3.Component(dna_identity, sbol3.SBO_DNA,
                        sequences=[dna_seq],
                        description=test_description)
if dna_comp.name == None:
    dna_comp.name = dna_identity
assert hl_dna_comp == dna_comp , f'Constructor Error: {dna_identity}'
```
Doing this I notice that the components are not equal, is this expected because they are indeed different components with the same information? How can I compare them for the test?

jakebeal: That's right: they are not equal because they are different objects. The sbol_diff package has functions for comparing objects, however, which you may find useful for this purpose.

Gonza10V: @jakebeal I updated the test as you required, please take a look at the `test_high_level_constructors` implementation using `doc_diff` which is the only tool that I found useful for this. It would be more useful for me to have a `toplvl_diff` or something like that, that sounds useful for other workflows?

sbol_utilities/component.py
```diff
@@ -246,6 +246,8 @@ def rna_component_with_sequence(identity: str, sequence: str, **kwargs)-> Tuple[
     rna_comp = sbol3.Component(identity, sbol3.SBO_RNA,
                                 sequences =[comp_seq],
                                 **kwargs)
+    if rna_comp.name == None:
+        rna_comp.name = identity
```
jakebeal, line 250: I don't think that you want to put a URI in as the names. "https://www.ncbi.nlm.nih.gov/nuccore/ATP07149_1" is not a very good name. I think you're looking for the `display_id`. I think it would be better, though, to just leave the name blank, as the spec says:

> If an Identified object lacks a name, then software tools SHOULD instead display the object’s displayId or URI.

If you set the name, then you are saying "this is the right name", if you leave it unset, then you are saying "the user didn't give a name, so this object should fail over to showing display_id"

jakebeal (changes requested):
The testing looks good!

I think you should roll back the change that sets name to identity, however.

test/test_component.py
jakebeal: doc_diff looks like it's working for you here. If you think toplvl_diff would be useful, please open an issue for it!

Gonza10V: This request is on issue #94 now

sbol_utilities/component.py
Gonza10V: > I don't think that you want to put a URI in as the names. "https://www.ncbi.nlm.nih.gov/nuccore/ATP07149_1" is not a very good name.

In ExternallyDefined objects the URI is the definition, the name can be setted using **kwargs if I understood the first part. The second argument is very good  so I rolled back the if statements that copies identity  on name.

jakebeal (approved):
Looks good! Merge when ready!

Gonza10V:
Thanks @jakebeal for your time and patience , I learned a lot making this contribution.

jakebeal:
FYI - you might be interested in some style-cleanup that I ended up doing this morning as a side effect of other work: #95.  If you're using an IDE like PyCharm, it can help with style compliance and type checking, which can be useful.

Issue #33, opened 2021-09-28
Set up read-the-docs


jakebeal:

noahsprent:
I've been working on this in a fork:

https://sbol-utilities-ns.readthedocs.io/en/latest/index.html

I've added what I think are the bare-minimum files for read-the-docs and sphinx for generating them. We just need a .readthedocs.yaml configuration file in the root of the repo and this points to a docs folder that contains the sphinx workings.

The only step after merging this will be to import the repo to read-the-docs, which maybe you might want to be the admin of @jakebeal ?

If I can be given access rights to the repo I can submit a pull request from my fork and assign this issue to myself.

jakebeal:
That sounds like a good plan. I looked to see if we could just have the organization be the owner, but I guess readthedocs doesn't support that yet.

jakebeal:
Looks like the build is running, but the documentation is empty: https://readthedocs.org/projects/sbol-utilities/

noahsprent:
That is what I mean by a 'minimal' setup. We can now edit the files in `docs` to create the docs!

jakebeal:
Excellent; I'm leaving this issue open until the minimal docstrings build is added there

noahsprent:
Docs are now live with autoapi reference page here: https://sbol-utilities.readthedocs.io/en/latest/

PR #78, opened 2022-01-18
Read the docs

noahsprent:
Addressing issue #33  I have added the bits needed in the repo to connect to a read-the-docs site. This is a minimal setup that we can build from in future.

- I added a .readthedocs.yaml configuration file in the root of the repository
- I added a docs directory and populated it with the sphinx defaults using sphinx-quickstart
- I added a few sphinx extensions and changed the theme to the read-the-docs theme

I have been able to set up read-the-docs in my fork, which you can see [here](https://sbol-utilities-ns.readthedocs.io/en/latest/). In the unforked repository someone (probably @jakebeal) will need to set up a read-the-docs account and link this repository to it ([This tutorial](https://docs.readthedocs.io/en/stable/tutorial/#sign-up-for-read-the-docs) starting from 'Sign up for read the docs'). If we want to build docs for develop, not just main, then they'll need to point the docs to develop in Admin>Advanced settings>Default branch.


jakebeal (approved):
Looks good to me; I'm going to try to set up on readthedocs now

PR #87, opened 2022-01-26
Gh action to lint python and auto fix issues

### LINKED ISSUE #68: Add python linter
We need to automate code quality checking. To do this, we should add a GitHub Action that runs [a python linter](https://books.agiliq.com/projects/essential-python-tools/en/latest/linters.html), probably pylint and/or flake8. This will likely throw up a bunch of errors that, in turn, also need resolution and/or known issue marking.

mohitdmak:
Hello @jakebeal , this PR hopefully fixes #68 

- This currently uses flake8 and black linters.
- In case of style issues, a new commit is automatically made fixing changes recommended by black

Do we need to proceed with recommended code styles by these linters, or:
- Lint code by customizing lint rules?


jakebeal:
Thank you for the contribution!

Three thoughts regarding your question:
1) I notice that flake8 is declaring a failure, but the code is passing in the action?
2) It looks to me like it's being a little bit heavy-handed on the formatting --- can we dial it down to be less sensitive (e.g., just PEP8 compliance)?
3) I'm not sure I'm comfortable doing auto-reformatting. I'd prefer to have it declare PEP8 violation errors and have the person pushing responsible for fixing.

mohitdmak:
Regarding 
1.  Actually, gh actions don't support auto commits or annotations on PRs from a forked repo, it is referenced by the creator of this action here: [link](https://github.com/wearerequired/lint-action#limitations) . I have tried the workaround mentioned [here](https://github.com/wearerequired/lint-action/issues/13) by triggering the actions on `pull_request_target` events which should allow annotations on non-local PRs.
2.  This should be possible, would look into it.
3. Sure, I disabled the auto committing, and once you run the workflows we can test if the annotations have started working here.

jakebeal:
I guess I don't know what you mean by an annotation --- I thought we would see a failure due to linting errors.

PR #89, opened 2022-01-27
Docstr coverage tests using Interrogate

### LINKED ISSUE #86: Add tests for docstring coverage
We can use something like https://pypi.org/project/docstr-coverage/ to warn us when we're missing docstrings for functions.

mohitdmak:
Hello @noahsprent , this tries to fix #86 

1. I have used [Interrogate](https://github.com/econchick/interrogate), which is much better than, and based on [docstr-coverage](https://pypi.org/project/docstr-coverage/)
2. The `tests/test_docstr_coverage.py` is an added test which uses a configuration defined at `pyproject.toml` and can be easily edited from there.
3. There are a few further options:
      -  Run Interrogate on every commit using [pre-commit hooks](https://pre-commit.com/)
      -  Create a [github actions](https://github.com/marketplace/actions/python-interrogate-check) workflow


noahsprent:
Awesome, thanks @mohitdmak! Tested running interrogate locally in a fresh venv with just interrogate installed and looks like it's working for me, and looks like from Github actions the interrogate dependency is being added and the .toml file is being 'prepared' fine. 

docstr_coverage

I'll take your word on Interrogate over docstr-coverage, I just did a quick Google to see what was out there. How did you select the options in the .toml file? Are they just the defaults?

I'm not massively au-fait with item 3, but looks to me based on the .git/hooks directory in the repo that we aren't currently using pre-commit hooks anywhere, so might be better just include this as part of the actions. Given that currently we're failing the test it might be good to have this as more of warning for now while we decide which files to ignore and which ones need improvement. 

Let's see what @jakebeal thinks.

mohitdmak:
> I'll take your word on Interrogate over docstr-coverage, I just did a quick Google to see what was out there. How did you select the options in the .toml file? Are they just the defaults?
Yep, Interrogate has a lot more features, users and its repo looks more active too. The options in .toml are currently the defaults except the verbosity. I dropped them all in so that we get all configurable options in one place.
> I'm not massively au-fait with item 3, but looks to me based on the .git/hooks directory in the repo that we aren't currently using pre-commit hooks anywhere, so might be better just include this as part of the actions. Given that currently we're failing the test it might be good to have this as more of warning for now while we decide which files to ignore and which ones need improvement.

Yeah sure, pre-commit hooks are not being used currently in this repo, so even I guess it should be left as another test in actions.

noahsprent:
> Yep, Interrogate has a lot more features, users and its repo looks more active too. The options in .toml are currently the defaults except the verbosity. I dropped them all in so that we get all configurable options in one place.

Cool, seems like a good strategy.

> Yeah sure, pre-commit hooks are not being used currently in this repo, so even I guess it should be left as another test in actions.

You ok to add that then when it looks like it's working we can merge?

jakebeal:
I'll endorse this approach. Please also open an issue for addressing the errors so that we can change back to strict once it's complete.

test/test_docstr_coverage.py
```diff
@@ -0,0 +1,29 @@
...
+    Tests each module, function, classes, methods for presence of docstrings
+"""
+import unittest
+from interrogate import coverage
+from interrogate import config
+
+"""
+    Setting up test config
+"""
+print("\nProject Root taken as:")
```
jakebeal, line 11: Please reorganize this as a standard unit test, more like the other test_X.py functions in the directory.
That will also allow you to mark the test as a "known failure" for us to resolve by adding the missing information.

jakebeal (changes requested):

test/test_docstr_coverage.py
mohitdmak: The latest commit does this, also there is a optional readme badge now, do let me know if it should be removed.
I have relaxed the test constraint for now to let it pass.

noahsprent: I like the badge! Is there a way to print the summary that I posted a photo of above so that it comes up during the tests? That way we can see where we're missing docstrings quickly? Or is this not possible if we format it as a unit test?

mohitdmak: > I like the badge! Is there a way to print the summary that I posted a photo of above so that it comes up during the tests? That way we can see where we're missing docstrings quickly? I'd have thought `cov.print_results` should do this, or is this not possible if we format it as a unit test?

Fixed this, turns out pytest only lets captured stdout content printed with a `-s` flag. I think the action works fine now.

noahsprent: Excellent, looks like it's working to me!

noahsprent (approved):

noahsprent (approved):
I can't comment on the nature of the code, but the outputs certainly seem like what we want to me. Thanks!

interrogate_badge.svg
```diff
@@ -0,0 +1,58 @@
+<svg width="140" height="20" viewBox="0 0 140 20" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" xmlns:serif="http://www.serif.com/" style="fill-rule:evenodd;clip-rule:evenodd;stroke-linejoin:round;stroke-miterlimit:2;">
```
jakebeal, line 1: I like the badge! However, it looks like this file is static, which means it won't update when the tests improve.

I suspect a dynamic badge tied to the develop branch can be implemented with the aid of GitHub artifacts (see: https://github.com/actions/upload-artifact). That may take a little while to figure out, though. Can I suggest that you remove the badge for now so that we can get this pull merged in. Then, if you're up for figuring out artifacts, see if you can set up another pull to add a dynamic badge?

jakebeal (changes requested):
Looking very close to me --- I just want to remove the static badge for now, then you can add a dynamic one in later if you want.

interrogate_badge.svg
mohitdmak: Sure, I have removed the badge, didn't notice it wasn't getting updated. 
I will look into the dynamic badge though.

jakebeal (approved):

Issue #83, opened 2022-01-18
SBOL3/2 conversion of orientations


jakebeal:
Conversion of `Location` object `orientation` fields between SBOL3 and SBOL2 does not work correctly currently, due to https://github.com/sboltools/sbolgraph/issues/16

A workaround has been tested in the iGEM distribution in https://github.com/iGEM-Engineering/iGEM-distribution/pull/150/commits/c1be61405426629fd2d59be3644dc25e52c42f05

This needs to be imported upstream to this library, cleaned up (orientations are currently remapped twice in the `convert3to2` function) and needs to have tests added to validate that the conversion is working correctly.

PR #90, opened 2022-01-27
added fix for conversion between SBOL2 and SBOL3

Akshat-Kumar-0610:
Hi @jakebeal, tried to work on the issue #83,

1. In the function convert3to2 I have removed the part where orientation were mapped again after conversion to SBOL2 document.
2. For the test for validation of orientation I have compaired orientation string to 'http://sbols.org/v3#inline' and if it  matches to the string test fails since it must not be present in an SBOL2 document.

### DIFF
--- a/test/test_conversion.py
+++ b/test/test_conversion.py
@@ -62,6 +62,10 @@ class Test2To3Conversion(unittest.TestCase):
         assert doc2.componentDefinitions[0].sequences[0] == 'https://synbiohub.org/public/igem/BBa_J23101_sequence'
         assert doc2.sequences[0].encoding == 'http://www.chem.qmul.ac.uk/iubmb/misc/naseq.html'
         assert doc2.sequences[0].elements == 'tttacagctagctcagtcctaggtattatgctagc'
+        for c in doc2.componentDefinitions:
+            for sa in c.sequenceAnnotations:
+                for loc in sa.locations:
+                    assert loc.orientation != 'http://sbols.org/v3#inline'
 
     def test_genbank_conversion(self):
         """Test ability to convert from SBOL3 to GenBank"""
--- a/test/test_conversion.py
+++ b/test/test_conversion.py
@@ -63,23 +63,33 @@ class Test2To3Conversion(unittest.TestCase):
         assert doc2.componentDefinitions[0].sequences[0] == 'https://synbiohub.org/public/igem/BBa_J23101_sequence'
         assert doc2.sequences[0].encoding == 'http://www.chem.qmul.ac.uk/iubmb/misc/naseq.html'
         assert doc2.sequences[0].elements == 'tttacagctagctcagtcctaggtattatgctagc'
+
+    def test_3to2_orientation_conversion(self):
+        """Test ability to convert orientation from SBOL3to SBOL2"""
+        # Get the SBOL3 test document
+        tmp_sub = copy_to_tmp(package=['iGEM_SBOL2_imports.nt'])
+        doc3 = sbol3.Document()
+        doc3.read(os.path.join(tmp_sub, 'iGEM_SBOL2_imports.nt'))
+
+        # Convert to SBOL2 and check contents
+        doc2 = convert3to2(doc3)
         # ids of location block containing orientation before conversion
         location_ids_sbol3 = []
-        def change_if_location(o):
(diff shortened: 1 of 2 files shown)


test/test_conversion.py
```diff
@@ -62,6 +62,10 @@ def test_3to2_conversion(self):
         assert doc2.componentDefinitions[0].sequences[0] == 'https://synbiohub.org/public/igem/BBa_J23101_sequence'
         assert doc2.sequences[0].encoding == 'http://www.chem.qmul.ac.uk/iubmb/misc/naseq.html'
         assert doc2.sequences[0].elements == 'tttacagctagctcagtcctaggtattatgctagc'
+        for c in doc2.componentDefinitions:
+            for sa in c.sequenceAnnotations:
+                for loc in sa.locations:
+                    assert loc.orientation != 'http://sbols.org/v3#inline'
```
jakebeal, line 68: Can we also have a positive check that we _do_ have some SBOL2 inline marks where expected? Orientation is optional, and so we need to make sure we're actually exercising the test - I don't know if the test file has an orientation in it at all.

jakebeal (changes requested):

jakebeal:
This is looking good; I'd like to see a strengthening of the test case, since I'm not sure if the test file has an orientation listed at all.

Akshat-Kumar-0610:
Hi @jakebeal, I have made few changes:
1. I have shifted conversion of orientation to traverse method for recursive traversal of objects.
2. for the tests part I have tried to compare identity of location block if it has orientation attribute before and after conversion.
3. for the actual testing test_files/BBa_J23101.nt file which is being used in test_3to2_conversion function does not contain location block with orientaion. For testing the changes I used test_files/iGEM_SBOL2_imports.nt file through python's interactive shell.

test/test_conversion.py
```diff
@@ -62,10 +63,23 @@ def test_3to2_conversion(self):
         assert doc2.componentDefinitions[0].sequences[0] == 'https://synbiohub.org/public/igem/BBa_J23101_sequence'
         assert doc2.sequences[0].encoding == 'http://www.chem.qmul.ac.uk/iubmb/misc/naseq.html'
         assert doc2.sequences[0].elements == 'tttacagctagctcagtcctaggtattatgctagc'
+        # ids of location block containing orientation before conversion
+        location_ids_sbol3 = []
+        def change_if_location(o):
```
jakebeal, line 68: This isn't actually changing locations: its name should describe what it actually does.

test/test_conversion.py
```diff
@@ -62,10 +63,23 @@ def test_3to2_conversion(self):
...
+            if isinstance(o, sbol3.Location):
+                if hasattr(o, 'orientation') and o.orientation != None:
+                    location_ids_sbol3.append(o.identity)
+        doc3.traverse(change_if_location)
```
jakebeal, line 72: How many should there be? Add a check to see the number.

jakebeal (changes requested):
Thank you, nice progress. Just a couple of small tweaks to request in the code. 

With regards to this:
> for the actual testing test_files/BBa_J23101.nt file which is being used in test_3to2_conversion function does not contain location block with orientation. For testing the changes I used test_files/iGEM_SBOL2_imports.nt file through python's interactive shell.

I'm glad that it worked for your local test. We won't know if it's still working once somebody else changes the code, however, unless we add a test to the test suite. Can you add a new test that uses `test_files/iGEM_SBOL2_imports.nt`? Tests are cheap and bugs are expensive!

Akshat-Kumar-0610:
Should I rewrite whole `test_3to2_conversion` function to use `test_files/iGEM_SBOL2_imports.nt` file or add another function just to test conversion of orientation?

jakebeal:
I would suggest adding another function to test conversion of orientation. The reason I suggest that is that one often gets better coverage and simpler debugging when there are a larger number of more focused tests, rather than a smaller number of more complex tests.

Akshat-Kumar-0610:
Hi @jakebeal  I have added separate test to validate orientation conversion using file `test_files/iGEM_SBOL2_imports.nt` and all 11 tests are passing as of now.

jakebeal (approved):

jakebeal:
Looks good --- nice work and thank you!

PR #92, opened 2022-01-31
Sep054

### LINKED ISSUE #62: Create a SEP 054 Package from a set of Modules
Just as a Module can be created from an SBOL Document (#60), a Package can be created from a set of Modules, following [SEP 054](https://github.com/SynBioDex/SEPs/blob/master/sep_054.md).

hgscott:
Resolves #62 

I changed the define_module and check_namespace functions to take *args, and so that they work on any number of sbol docs provided. 

I updated test_package to have a second test function, test_define_package_multi_file, that takes 3 (very contrived) sbol documents and creates one package file.

I do want to check about how to properly concatenate files. In lines 101 and 102 of package.py, I am combining the files by adding all objects in one file to another.


sbol_utilities/package.py
```diff
@@ -64,26 +64,27 @@ def regularize_package_directory(dir: str):
         raise ValueError(f'Package {dir}: {PACKAGE_DIRECTORY} subdirectory should not have any subdirectories of its '
                          f'own, but found {package_sub_dirs[0]}')
 
-def define_package(doc: sbol3.Document):
-    """Function to take an sbol document, check if it is a package, and create
-    a new sbol document with the package definition included
+def define_package(*args: sbol3.Document):
```
jakebeal, line 67: I believe you don't need to call the variable args --- you could use call it *docs instead
See: https://stackoverflow.com/questions/53676249/are-python-args-and-kwargs-ever-named-something-else-in-practice

sbol_utilities/package.py
```diff
@@ -94,26 +95,33 @@ def define_package(doc: sbol3.Document):
         # Namespace must match the hasNamespace value for all of its members
         package.namespace = package_namespace
 
+        # Combine all of the docs into one
+        doc = args[0]
+        for sbol_doc in args[1:]:
+            doc_objects = [o for o in sbol_doc.objects]
```
jakebeal, line 101: Copying the objects is making a build artifact for the package (see [this section in SEP 054](https://github.com/SynBioDex/SEPs/blob/master/sep_054.md#collating-a-native-package-build-artifact)), which we will also want to have a function for - you can break this out into that function.  

For defining a package, we just need to take the inventory of each of the sub-package documents.

sbol_utilities/package.py
```diff
@@ -94,26 +95,33 @@ def define_package(doc: sbol3.Document):
...
+            doc.add(doc_objects)
+        
         # Add the package to the document
         doc.add(package)
```
jakebeal, line 105: On reflection, I don't think we actually need or want to add the package automatically. We can build the package object without adding it to anything, then add it later (or in a convenience wrapper function, like this pass-through adder in component.py: https://github.com/SynBioDex/SBOL-utilities/blob/3fa0f3a2bac2833f82c486a6cfb0d66660963ceb/sbol_utilities/component.py#L80

sbol_utilities/package.py
```diff
@@ -94,26 +95,33 @@ def define_package(doc: sbol3.Document):
...
 
         # Return the doc
         return(doc)
         
 
-def check_namespaces(doc: sbol3.Document):
-    """ Check if the namespaces of all top level objects are the same
+def check_namespaces(*args: sbol3.Document):
```
jakebeal, line 111: I think we should change this to apply to the package object, rather than the document. The approach we've embraced in pySBOL is "build, then check".

sbol_utilities/package.py
jakebeal: I also think it will be important to distinguish between the first document, which will be for the package, and the rest of the documents, which will be for sub-packages.

jakebeal (changes requested):
Good progress; I have some suggestions for improvement.

sbol_utilities/package.py
hgscott: I have updated the check_namespaces function to work on a package, it now checks the namespaces of all of the members of the package, checks that all sub-packages have a namespace with a stem of the namespace of the root package, and then checks that each subpackage has members of a single namespace

sbol_utilities/package.py
hgscott: I am now passing the first document and the sub-package documents separately, but I immediately combine them all into one list and treat them all the same. When aggregating multiple files, do I need to be treating the root package file differently?

sbol_utilities/package.py
hgscott: I wasn't really sure what this section in the SEP markdown meant- but I am no longer combining files in this way.

hgscott:
Thanks for your comments Jake- I made some changes but was still a little fuzzy on the final structure of the resulting packages and subpackages

sbol_utilities/package.py
```diff
@@ -64,107 +65,186 @@ def regularize_package_directory(dir: str):
...
 
     Returns:
         is_package (boolean): True if all namespaces are the same
-        namespace (string): Namespace of the package
     """
-
+    # Check all members of the root package have the same namespace
     # Get a list of all namespaces
-    all_namespaces = [o.namespace for o in doc.objects]
+    # To get a namespace, remove the object name from the URI for each member
```
jakebeal, line 202: I believe this should still be using the o.namespace property.

sbol_utilities/package.py
```diff
@@ -64,107 +65,186 @@ def regularize_package_directory(dir: str):
...
+    all_namespaces = ["/".join(URI.split('/')[0:-1]) for URI in package.members]
 
     # Check all namespaces are the same
-    is_package = all(x == all_namespaces[0] for x in all_namespaces)
-
-    # Get the first namespace to pass the string
-    # Which position you pick won't matter if it is a package
-    namespace = all_namespaces[0]
+    if all(x == all_namespaces[0] for x in all_namespaces):
```
jakebeal, line 206: Note that another, slightly more elegant way to do this, is to have all_namespaces be generated as a set, rather than a list. Using a set will automatically prune duplicates. Then you can just check if the len(all_namespaces)==1.

sbol_utilities/sep_054_extension.ttl
```diff
@@ -79,35 +79,29 @@ sep054:includeObjectReferences rdf:type owl:DatatypeProperty ;
...
-        rdfs:comment '''This property links to a description of a package dependency''' ;
-               rdfs:subPropertyOf sbol:directlyComprises ;
-              rdfs:domain sep054:Module ;
-              rdfs:range sep054:Dependency ;
-              rdfs:label "dependencies" .
-
-
-sep054:Package rdf:type owl:Class ;
-        rdfs:comment  '''A Package is a systematic aggregation of modules intended to be distributed and used as a
-        coherent whole. When stored in a hierarchical structure (e.g., files in directories), the relationship between
```
jakebeal, line 100: I believe that the comment about hierarchical directory structure should be kept.

test/test_files/package_out_02.nt
```diff
@@ -0,0 +1,9 @@
+
+<https://example.org/MyPackage/package> <http://sbols.org/SEP054#conversion> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
+<https://example.org/MyPackage/package> <http://sbols.org/SEP054#subPackage> <https://example.org/MyPackage/package> .
```
jakebeal, line 3: Shouldn't have itself as a sub-package. Instead, it should have all of the same contents as package_out_01.nt, but with the addition of the two subpackages.

jakebeal (changes requested):
Coming along well; I noted a couple of bugs, but most of my comments are about style/elegance and should be considered optional.

sbol_utilities/package.py
jakebeal: Yes, this was creating a bug. I've added an explanatory comment both in the code here and in the relevant .nt file, which illustrates the problem well. Right now, you add the package as a sub-package of itself, which both creates a loop and also doesn't store the members.

sbol_utilities/sep_054_extension.ttl
hgscott: I added the comment back in at the bottom of the rds comment (lines 95-97).

sbol_utilities/package.py
hgscott: I am unsure how to do this- when working on a pre-defined package the way I am accessing the "objects" is through the `members` attribute of the packages. But `members` is a list of URI strings, and don't have attributes like `namespace`.

sbol_utilities/package.py
```diff
@@ -64,107 +65,186 @@ def regularize_package_directory(dir: str):
...
-        doc (sbol3.Document): The document with the added package info
+        package: The package definition
     """
-    # Call check_namespace
-    logging.info('Checking namespaces')
-    is_package, package_namespace = check_namespaces(doc)
-    logging.info(f'SBOL Document is a package: {is_package}')
+    # Put all the input files in one list
+    # Means I am treating the root and subpackage files exactly the same
+    # CHECK: How should I be treating the root and the sub packages differently?
```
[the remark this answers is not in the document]
jakebeal: Hmm... that probably does need clarification. It's specifically talking about the case when the package is a directory. In that case, it wouldn't have members because there's not file, just a directory containing sub-package files. Do you want to suggest a change and have me review, or vice versa?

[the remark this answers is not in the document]
hgscott: I think it would be best if you made the change- I feel like still do not have a firm grasp on working with a directory rather than with files.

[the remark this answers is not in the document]
jakebeal: Made the change: https://github.com/SynBioDex/SEPs/commit/5faba2733ae87c83ebe09fb70c3248da2f7583cf
Please let me know if that makes more sense to you.

sbol_utilities/package.py
```diff
@@ -64,107 +65,186 @@ def regularize_package_directory(dir: str):
...
-    # If all namespaces are the same, add package and save as new file
-    if is_package:
+    Args:
+        package_file (sbol3.Document): SBOL document containing a package
+
+    Return
+        package: The package definition
+    """
+    # Get the namespace for the first object, if the file is a package, all 
+    # namespaces should be the same, so which one you pick does not matter
```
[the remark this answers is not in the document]
hgscott: Added the check with a set in commit hgscott/SBOL-utilities@312854b3969d46a59f34f4ee11cd5bd2c352da0b

sbol_utilities/package.py
hgscott: Added the check with a set in commit hgscott/SBOL-utilities@312854b3969d46a59f34f4ee11cd5bd2c352da0b

jakebeal (approved):
I can see some future refinements, but this looks good enough for the first merge.

Issue #69, opened 2022-01-17
Function to check if a part is a composite


jakebeal:
Following [SEP 055](https://github.com/SynBioDex/SEPs/blob/master/sep_055.md), a composite part can be detected by checking if the part was generated by an assembly plan.

Add a function that looks at a component to determine if it is a composite part.

PR #98, opened 2022-02-03
added helper function to check composite

Akshat-Kumar-0610:
I have  tried to work on issue #69.
I have added a helper function `is_composite` in `sbol_utilities/helper_functions.py`.
I have added checks for component, dna type and has generated_by attribute.


sbol_utilities/helper_functions.py
```diff
@@ -133,3 +133,25 @@ def has_plasmid_role(x):
...
+def is_composite(obj: sbol3.Component) -> bool:
+    """Check if an SBOL Component is a composite.
+
+    :param obj: design to be checked
+    :return: true if composite
+    """
+    def has_dna_type(o):
+        for t in o.types:
+            try:
+                regularized = tyto.SBO.get_uri_by_term(tyto.SBO.get_term_by_uri(t))
```
jakebeal, line 146: I think this can be done more compactly (and more safely) using tyto's comparison tests.

sbol_utilities/helper_functions.py
```diff
@@ -133,3 +133,25 @@ def has_plasmid_role(x):
...
+                if regularized == tyto.SBO.deoxyribonucleic_acid:
+                    return True
+            except LookupError:
+                pass
+        return False
+
+    if isinstance(obj, sbol3.Component):  # Composite is SBOL component
+        if has_dna_type(obj):  # Composite has type SBO:DNA
+            if hasattr(obj, "generated_by") and len(obj.generated_by) != 0: # Composite has attribute prov:generated_by linking to prov:activity
```
jakebeal, line 155: I think that checking for an SEP 054 assembly plan will need to have a more specific inspection of what sort of generation has occurred.

jakebeal (changes requested):
Good start. I added a couple of things I see that need improvement.
This also needs tests to check if it is working.

sbol_utilities/helper_functions.py
Akshat-Kumar-0610: Can you please elaborate on this. Is this a method defined in tyto's library?

jakebeal: Check out tyto's inference methods: https://tyto.readthedocs.io/en/latest/index.html#inference

sbol_utilities/helper_functions.py
Akshat-Kumar-0610: I tried to check for entries in `generated_by` attribute whether they are assembly plan or not, but none of the files in the `test_files/` directory seems to contain a proper assembly plan. I tried checking wrt this point `Specifically, the Component for the composite part SHOULD have a prov:wasGeneratedBy propert whose value is a prov:Activity object with two values of its type property, one being sbol:design and the other being the new term http://sbols.org/v3#assemblyPlan`. can you please confirm whether this observation is correct or am I missing something.

jakebeal: That is correct: all of the current test files were written before SEP 055 was drafted, and so don't contain an assembly plan in its form. You would want to create an example to check.

jakebeal:
No activity has happened on this pull request for more than a year, so I am closing it.

Issue #96, opened 2022-02-03
Move docstring interrogate to a separate test job


jakebeal:
Right now, interrogate is run as part of the cross-platform matrix
It would be better to run as a separate job, not mixed in with the other tests.

PR #99, opened 2022-02-03
Docstr Interrogate in a seperate job

mohitdmak:
Hey @jakebeal , for #96 , I have moved docstr_coverage testing before the build and other tests on all platforms.


jakebeal (approved):
Looking good. I'll bet the docstr_coverage job can be used to produce a badge as a build artifact too.

mohitdmak:
> Looking good. I'll bet the docstr_coverage job can be used to produce a badge as a build artifact too.

I have added the badge by having Docstrings check as a new workflow, is that ok as we are also going to have another workflow for linting?
Also currently a post request to "validator_url" in a few tests is not getting any response causing them to fail.

jakebeal:
I think using the workflow badge is a reasonable solution. It would be a bonus we could get the version with the number in there, but that's a nice-to-have. Do please switch the link to be this repository rather than yours.

The test failures are because https://validator.sbolstandard.org/ is currently down due to a certificate issue; the maintainers are trying to bring it back up again right now.

README.md
```diff
@@ -2,6 +2,7 @@
 SBOL-utilities is a collection of scripts and functions for manipulating SBOL 3 data that can be run from the command line or as functions in Python.
 
 [![Documentation Status](https://readthedocs.org/projects/sbol-utilities/badge/?version=latest)](http://sbol-utilities.readthedocs.io/)
+[![Docstrings Coverage](https://github.com/mohitdmak/SBOL-utilities/actions/workflows/docstr-coverage.yml/badge.svg)](https://github.com/mohitdmak/SBOL-utilities/actions/workflows/docstr-coverage.yml)
```
jakebeal, line 5: Please change link location.

jakebeal (changes requested):

jakebeal:
Validator service is up again, so tests are working.

README.md
mohitdmak: Sure

jakebeal:
It looks like your interrogate run is having an issue due to the relative paths. You may want to update from develop, as I ran into this issue also and put in a patch to use a path based on file location. See: https://github.com/SynBioDex/SBOL-utilities/commit/dc9b609c944172fd25f521b2db1ba4b1b1735eed#diff-21a6bd99abbaa8a90bda9afdeaf8c25f2a16b792fa62d29d23ef13083e478e51

jakebeal:
I think this rabbit hole is getting too deep... let's just fall back to a pass/fail badge.

jakebeal:
Please let me know when you're ready for another review.

mohitdmak:
> Please let me know when you're ready for another review.

Yep, I just reverted to our previous pass/fail badge and fixed its link to your this repo instead of my fork, I guess it can be merged if we are finally sticking to pass/fail?

docstr_badge.svg
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jakebeal, line 1: With the pass/fail badge plan, I believe this file isn't being referenced any more. Please delete it, and then we're good for merge.

jakebeal (changes requested):

docstr_badge.svg
mohitdmak: Oh, my bad, I have got it deleted now.

jakebeal (approved):

Issue #93, opened 2022-01-31
Duplication of graph_sbol utility in the root directory


karin-sionov:
Currently there are two scripts that are performing the same thing: 
- https://github.com/SynBioDex/SBOL-utilities/blob/develop/graph-sbol (in the root directory) 
- https://github.com/SynBioDex/SBOL-utilities/blob/develop/sbol_utilities/graph_sbol.py (in the scripts folder) 

The first one (in the root dir) should be deleted and replaced with a call to graph_sbol.py from setup.py. 
In order to remove this redundancy, we need to: 
- compare the files (make sure we are not losing data when removing one of them) 
- in `setup.py`, replace the `scripts=['graph-sbol']` with a call to the functions within sbol_utilities/graph_sbol.py. 
- make sure everything still works
- delete redundant `graph-sbol`

Issue #91, opened 2022-01-31
Fix inconsistency of utilities names in the documentation 


karin-sionov:
This is a documentation issue. 
In the README file, the name of the utilities does not always match the name of the actual python script. 
For example, the utility `sbol-converter` is actually being handled by the script `conversation.py`.  
I think that the `sbol-calculate-sequences` utility refers to the script `component.py`. 

So it is either changing the script name, or its reference in the documentation - @jakebeal is there a guideline or something? 
I think it would be better to change the script name, if we want to keep the utilities names consistent with the "SBOL-" prefix. But it could be ugly to change a script name, so whatever you prefer...

jakebeal:
This is a good question, @karin-sionov . The names of the command-line utilities are being set in setup.py, which gives the names for "wrapper" scripts that call functions in packages.  

For example, sbol-converter is defined on this line as a call to the main function in sbol_utilities.conversion:main. The next few lines define the other converter utilities as calls to other entry-point functions in the same Python module.
https://github.com/SynBioDex/SBOL-utilities/blob/3fa0f3a2bac2833f82c486a6cfb0d66660963ceb/setup.py#L48-L54

Clearly the either the current naming scheme or its documentation could be improved. What do you think would be a good approach?

ahmedtarek26:
@karin-sionov, @jakebeal some utility names are descriptive like `sbol2to3` but we can name it `sbol2-to-sbol3` to make it more clear, others like `sbol2fasta` may be confusing because we have sbol2, so I suggest making it `sbol-to-fasta`

jakebeal:
@ahmedtarek26 That sounds like a reasonable proposal. Want to set up a pull request for that change?

ahmedtarek26:
@karin-sionov, @jakebeal I did a pull request for that changes, have a look at it, and if there are more suggestions I'm waiting to hear them.

fahdalkhulaifi:
Hi @jakebeal 👋

I've been looking into this issue and noticed that while @ahmedtarek26 opened a PR in 2023 to align utility names, the script names under `sbol_utilities/` haven't changed since then, and discussion has been quiet.

Rather than renaming scripts (which might break imports), would it make sense to update the README to clearly map CLI commands like `sbol-converter` to their corresponding functions in `conversion.py`, etc.?

That way, the documentation is accurate without introducing any breaking changes.

Happy to open a PR if that sounds like a helpful direction!

jakebeal:
[...]

PR #101, opened 2022-02-04
Switch to entrypoint console_script for graph-sbol

mohitdmak:
Hey @jakebeal  @karin-sionov for #93 ,
( Currently the tests are failing due the same reason as [this](https://github.com/SynBioDex/SBOL-utilities/pull/99#issuecomment-1029593464))
- I have removed the executable of graph-sbol and added a console_script in `setup.py` for graph-sbol which uses `sbol_utilities/graph_sbol.py`. 
- I think this is better than having a "graph-sbol" executable declared in script section of `setup.py`, due to [these reasons](https://stackoverflow.com/questions/18787036/difference-between-entry-points-console-scripts-and-scripts-in-setup-py)
- But I am having trouble getting a paml file to test the graph-sbol function with


karin-sionov:
Hi @mohitdmak , thank you for handeling the issue. I agree that the console is more elegant than using a script declaration. 
Regarding the PAML file, are you using this package? [PAML Github](https://github.com/Bioprotocols/paml)

jakebeal (changes requested):
This code all looks good, thank you!

The test failures are because https://validator.sbolstandard.org/ is currently down due to a certificate issue; the maintainers are trying to bring it back up again right now.

For adding tests, graph-sbol should not need PAML: it should be able to run on any SBOL file. I would suggest using some of the files in `test/test_files`. If would likely be good to try both a simple one so that you know exactly what it should look like and one of the more complex ones, like constraints_library.nt

jakebeal:
Validator service is up again, so tests are working.

mohitdmak:
> For adding tests, graph-sbol should not need PAML: it should be able to run on any SBOL file. I would suggest using some of the files in `test/test_files`. If would likely be good to try both a simple one so that you know exactly what it should look like and one of the more complex ones, like constraints_library.nt

Oh thanks, I tried calling the main() in `graph_sbol.py` along with the sbol file, and it properly creates a pdf, so I guess the console_script should also work.

- Also, @jakebeal  would we need any sort of test for this utility? (maybe having the same approach as other tests like having graphed PDFs of a sample sbol file, and creating another one during the test, and using doc_diff to check if they are the same)
- And about #91 , since its related to setup.py scripts too, how about just naming scripts with "sbol-" prefix and not changing names of the py files themselves, as they are imported at many places?

jakebeal:
@mohitdmak I have a mantra for code development: anything not pushed to the repository doesn't exist and anything not tested in the CI is broken. :-)

Comparing PDFs is problematic, though, since they typically contain things like datestamps that mean two generated PDFs are never exactly the same.  I think, however, that this suggests a simple new feature that will also help us test. Right now, the intermediate graphviz file is never produced, only the PDF that comes from it. Graphviz can also produce other formats, however, including readily diffed formats like SVG.  Want to add an "output format" argument that merely defaults to PDF, but can also be set to graphviz (.dot) and other output formats? See: https://graphviz.org/docs/outputs/

Re #91: I'd suggest discussing that with @karin-sionov and anyone else interested on that issue, then handling it with a separate pull request.

mohitdmak:
> @mohitdmak I have a mantra for code development: anything not pushed to the repository doesn't exist and anything not tested in the CI is broken. :-)

Yep, sure got it.

> Comparing PDFs is problematic, though, since they typically contain things like datestamps that mean two generated PDFs are never exactly the same. I think, however, that this suggests a simple new feature that will also help us test. Right now, the intermediate graphviz file is never produced, only the PDF that comes from it. Graphviz can also produce other formats, however, including readily diffed formats like SVG. Want to add an "output format" argument that merely defaults to PDF, but can also be set to graphviz (.dot) and other output formats? See: https://graphviz.org/docs/outputs/

1. This worked very well! I have added a unit test comparing `.dot` files, but for it, I have also modified the `graph_sbol.py` itself, to have additional parser arguments for: 
    - **file_format** (defaults to pdf for users invoking it from CLI, and we can specify "dot" format for our tests),
    - **view_now** (defaults to not active again, and we can specify the flag during the unit test to not automatically open generated file)

2. So currently it tests only the `BBa_J23101.nt` file by generating a `BBa_J23101_test.dot`, and then we assert it to be identical to an existing `BBa_J23101.dot` file. 

3. In the process an intermediate `BBa_J23101_test` file is also generated when `graphviz.Digraph()` is called, so after the test succeeds, I delete it along with it's .dot version. But if the test fails, those 2 generated files would not be deleted, allowing us to debug the graph-sbol utility.
Is this approach okay?

4. Also do we need to test more sbol files in `test_files` or just the one suffices?

> Re #91: I'd suggest discussing that with @karin-sionov and anyone else interested on that issue, then handling it with a separate pull request.

Sure, I've had a go here: #105

.github/workflows/python-app.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install pytest
         python -m pip install interrogate
+        sudo apt install graphviz
```
jakebeal, line 39: Note that an alternate option for installing graphviz is to have a step that uses the workflow that's been set up for doing so:
```
    - name: Setup Graphviz
      uses: ts-graphviz/setup-graphviz@v1
```

sbol_utilities/graph_sbol.py
```diff
@@ -4,7 +4,7 @@
 import argparse
 
 
-def graph_sbol(doc, outfile='out'):
+def graph_sbol(doc: sbol3.Document, file_format: str, view_now: bool, outfile: str ='out'):
```
jakebeal, line 7: I believe that file_format and view_now should have defaults.
I'd suggest view_now defaults to false (matching the -v to enable below) and format defaults to PDF

jakebeal (changes requested):
Just a very minor change remaining, I think

.github/workflows/python-app.yml
mohitdmak: Okay, I have replaced it.

sbol_utilities/graph_sbol.py
mohitdmak: Alright

jakebeal (approved):

PR #105, opened 2022-02-06
Setup.py changes in scripts

mohitdmak:
Hey @jakebeal , as discussed in the [2nd point here](https://github.com/SynBioDex/SBOL-utilities/pull/101#issuecomment-1030500850),  I have tried making those changes. I am not sure if we should be renaming any functions or file names in utilities themselves, should this be all we do so that at least the convention for scripts is fixed?


jakebeal:
@karin-sionov: I'd like to get your thoughts on naming here.

karin-sionov:
Hi, 
Few inputs: 
1. I think `sbol-excel-to-sbol` is a bit confusing name. Maybe just `sbol-convert-from-excel`? Or perhaps not adding the "sbol-" prefix in here? The same goes for `sbol-sbolToGenbank` and `sbol-sbolToFasta`: consider changing them to `sbol-convert-to-fasta` or something like that. 
2. It still does not match the Readme file, but that's an easy change I can make nyself once we decide upon the utilities names. 
3. I am still more bothered by the `sbol-converter` utility that remains `conversion` in the script... This is why I opened the #91  issue in the first place - seems odd to me to change all the utilities aside from that one. I know @mohitdmak suggested not to change it, but I admit I did not understand the reason, since in this commit you change all its inner functions so it looks like it requires the same amount of work :)

karin-sionov (commented):
I think sbol-excel-to-sbol is a bit confusing name. Maybe just sbol-convert-from-excel? Or perhaps not adding the "sbol-" prefix in here? The same goes for sbol-sbolToGenbank and sbol-sbolToFasta: consider changing them to sbol-convert-to-fasta or something like that.
It still does not match the Readme file, but that's an easy change I can make nyself once we decide upon the utilities names.
I am still more bothered by the sbol-converter utility that remains conversion in the script... This is why I opened the https://github.com/SynBioDex/SBOL-utilities/issues/91 issue in the first place - seems odd to me to change all the utilities aside from that one. I know @mohitdmak suggested not to change it, but I admit I did not understand the reason, since in this commit you change all its inner functions so it looks like it requires the same amount of work :)

karin-sionov:
Actually I take my 3rd point back. I think calling the main utility "SBOL-converter" is enough without changing the script's name (.py), since the script also calls other utilities. 
So @mohitdmak what do you think about the first two points?

mohitdmak:
Hi @karin-sionov ,

1. Yes, actually regarding the 3rd point, I was a bit averse to changing the script's name as it would involve changing the script's reference in multiple places, and it might be better to just have the final executables have proper naming. So I agree with your latter comment in this regard.
2. For the 1st point, I don't have any problem following your naming suggestion, it seems more intuitive than mine.
3. As for the 2nd point, I think I changed the readme for the changed console scripts in the current commit. Is any of them left out?

jakebeal:
This doesn't seem to have been needed, and we are declining to make the change.

PR #109, opened 2022-02-09
add test for design_file_type with GenBank extension and full path support

karin-sionov:
I noticed that GenBank extension support is not being tested, and also the full path support is not tested. So I added 1 test to cover both.


jakebeal (approved):
Nice catch, thank you! Looks like things still pass, so that is good.

PR #111, opened 2022-02-12
Add filter by roles functionality

mohitdmak:
This adds the ability to filter a document's components by roles, as discussed in [this issue](https://github.com/SynBioDex/pySBOL3/issues/204).
Also added a test for the same.

### DIFF
--- a/test/test_component.py
+++ b/test/test_component.py
@@ -8,7 +8,7 @@ import sbol3
 import tyto
 
 from sbol_utilities.component import contained_components, contains, add_feature, add_interaction, constitutive, \
-    regulate, order, in_role, all_in_role, ensure_singleton_feature, by_roles, by_types, is_dna_part, ed_restriction_enzyme,
+    regulate, order, in_role, all_in_role, ensure_singleton_feature, by_roles, by_types, is_dna_part, ed_restriction_enzyme
 from sbol_utilities.helper_functions import filter_top_level
 from sbol_utilities.component import dna_component_with_sequence, rna_component_with_sequence, \
     protein_component_with_sequence, media, functional_component, promoter, rbs, cds, terminator, \


mohitdmak:
@jakebeal  Can you review this too?

jakebeal:
Apologies; this slipped by my notice. I will review it now.

sbol_utilities/component.py
```diff
@@ -54,6 +54,21 @@ def check_roles(component: sbol3.Component) -> bool:
...
+    :param doc: sbol3 document to search in
+    :param required_role: the role which must be present in components
+    :return: list of matched components
+    """
+    # callable to be provided as an arg to find_all() of Document class
+    def has_role(obj: sbol3.Identified) -> bool:
+        return isinstance(obj, sbol3.Component) and required_role in obj.roles
+
+    # search through all Identified objects
+    return doc.find_all(has_role)
```
jakebeal, line 69: I think this can be done more parsimoniously by calling the new `filter_top_level` function that you just added.
While doing this, you might also want to add a complementary `filter_by_type` function.
If both of these return iterables (per `filter_top_level`), then they can be chained together as well.

jakebeal (changes requested):
I think this can be raised a level in functionality by taking advantage of the new `filter_top_level` function that you just added.

sbol_utilities/component.py
mohitdmak: I have tried using `filter_top_level` here, but the `by_roles`, and `by_types` functions are reduced just to a helper function for `filter_top_level`. I am not sure if this is what you meant, can you look at the unit test functions and see if they use these functions in a good way?

sbol_utilities/helper_functions.py
```diff
@@ -80,15 +80,18 @@ def toplevel_named(doc: sbol3.Document, name: str) -> Optional[sbol3.Identified]
...
+def filter_top_level(doc: sbol3.Document, filter: Callable[[sbol3.TopLevel], bool], **kwargs) -> Iterable[sbol3.TopLevel]:
     """Filters and returns iterable of TopLevel Objects in a document,
     which match a criteria set by a callable argument.
 
     :param doc: SBOL Document to search
     :param filter: Callable acting as filter on List of TopLevel objects
     :return: TopLevel iterator satisfying given filter
     """
-    return (obj for obj in doc.objects if filter(obj))
+    if kwargs:
```
jakebeal, line 91: Is this if statement necessary? I think the two branches are equivalent.

test/test_component.py
```diff
@@ -17,6 +18,38 @@
...
+        # create and add 3 components, with 2 having common role of dna
+        comp_1 = sbol3.Component('component_1', sbol3.SBO_DNA, roles=[tyto.SBO.deoxyribonucleic_acid])
+        comp_2 = sbol3.Component('component_2', sbol3.SBO_DNA, roles=[tyto.SO.engineered_region])
+        comp_3 = sbol3.Component('component_3', sbol3.SBO_DNA, roles=[tyto.SO.engineered_region, tyto.SBO.deoxyribonucleic_acid])
+        doc.add(comp_1)
+        doc.add(comp_2)
+        doc.add(comp_3)
+        # only comp_1 and comp_3 must be returned by the function
+        matched_itr = filter_top_level(doc, by_roles, required_role=tyto.SBO.deoxyribonucleic_acid)
+        matched = [obj for obj in matched_itr]
```
jakebeal, line 34: You can write this more simply as just `matched = list(filter_top_level... `
(and below as well)

jakebeal (changes requested):
Just a couple of places where I think the code could use small improvements.

sbol_utilities/component.py
```diff
@@ -60,6 +60,26 @@ def check_roles(component: sbol3.Component) -> bool:
         and has_dna_type(obj) and len(obj.sequences) == 1
 
 
+def by_roles(obj: sbol3.TopLevel, required_role: str) -> bool:
+    """Search the entire provided document, and return components which have specified role as one of its roles.
```
jakebeal, line 64: This description isn't quite correct, nor the one for by_types

sbol_utilities/component.py
```diff
@@ -60,6 +60,26 @@ def check_roles(component: sbol3.Component) -> bool:
         and has_dna_type(obj) and len(obj.sequences) == 1
 
 
+def by_roles(obj: sbol3.TopLevel, required_role: str) -> bool:
```
jakebeal, line 63: This can be cleaner if instead of returning a boolean, you return a lambda that returns a boolean. Then you can construct the function and pass the function into filter_top_level rather than passing its arguments into filter_top_level

jakebeal (changes requested):

sbol_utilities/component.py
jakebeal: Having simpler helper functions is fine. As I note below, I think it will do better if they return lambdas, however.

sbol_utilities/helper_functions.py
mohitdmak: Fixed, I had thought initially that in absence of keyword arguments, kwargs would return an empty dict causing an extra arg to be passed to the helper functions.

test/test_component.py
mohitdmak: Fixed

sbol_utilities/component.py
mohitdmak: Fixed, I forgot to change these while changing functions

jakebeal (approved):

tcmitchell:
@jakebeal you approved this on Feb 27, but it did not get merged. What is the status? Should the merge conflicts be resolved so that this can be re-reviewed and then merged?

jakebeal:
Yes, that would be a good idea.

mohitdmak:
@tcmitchell  @jakebeal  the merge is resolved, this can be merged now

Issue #71, opened 2022-01-17
Filter to collect TopLevel objects


jakebeal:
Add a macro/wrapper function that returns an iterable of all TopLevel objects that match one or more Boolean test criteria.  For example, something to get all of the parts in a document (ref: #70)

PR #115, opened 2022-02-22
Filter top level objects by any criteria

mohitdmak:
Hi @jakebeal , as a follow-up to #108 , I have tried #71.
It returns an iterable of top-level objects, and the test uses the newly added check in #108 on the `simple_library.nt` test file.

### DIFF
--- a/sbol_utilities/helper_functions.py
+++ b/sbol_utilities/helper_functions.py
@@ -1,6 +1,6 @@
 import logging
 import itertools
-from typing import Iterable, Union, Optional
+from typing import Iterable, Union, Optional, Callable, List
 
 import sbol3
 from sbol3.refobj_property import ReferencedURI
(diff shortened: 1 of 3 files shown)


sbol_utilities/helper_functions.py
```diff
@@ -80,6 +80,18 @@ def toplevel_named(doc: sbol3.Document, name: str) -> Optional[sbol3.Identified]
...
 
+def filter_top_level(doc: sbol3.Document, filter: Callable[[sbol3.TopLevel], bool]) -> Iterable[sbol3.TopLevel]:
+    """Filters and returns iterable of TopLevel Objects in a document,
+    which match a criteria set by a callable argument.
+
+    :param doc: SBOL Document to search
+    :param filter: Callable acting as filter on List of TopLevel objects
+    :return: TopLevel iterator satisfying given filter
+    """
+    result: List[sbol3.TopLevel] = [obj for obj in doc.objects if filter(obj)]
```
jakebeal, line 91: Rather than making a list and turning it into an iterable, you can return an iterable direcly by using a generator () instead of a list []

jakebeal (changes requested):
Looking good; just one small optimization I'd suggest.

sbol_utilities/helper_functions.py
mohitdmak: Oh alright.

PR #116, opened 2022-02-22
Add build_reference_cache to accelerate lookup

tcmitchell:
Look up of referenced objects is slow in large documents. Accelerate the lookup through the use of a cache. Add `build_reference_cache` to build the cache, and accept the cache as an optional argument to `find_child` and `find_top_level`.

Use the cache in `compute_sequence` and `calculate_sequences`.

### DIFF
--- a/sbol_utilities/helper_functions.py
+++ b/sbol_utilities/helper_functions.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import logging
 import itertools
 from typing import Iterable, Union, Optional, Callable, List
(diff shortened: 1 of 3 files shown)


sbol_utilities/helper_functions.py
```diff
@@ -37,26 +38,52 @@ def id_sort(i: iter):
...
+def find_child(ref: ReferencedURI, cache: Optional[dict[str, sbol3.Identified]] = None):
     """Look up a child object; if it is not found, raise an exception
 
     :param ref: reference to look up
     :returns: object pointed to by reference
     :raises ChildNotFound: if object cannot be retrieved
     """
+    try:
+        return cache[str(ref)]
+    except KeyError:
```
jakebeal, line 64: These exceptions could use a comment explaining why we're passing on them. Same below in find_top_level.

jakebeal (approved):

sbol_utilities/helper_functions.py
tcmitchell: @jakebeal This is a great point. I added comments explaining what the raised exceptions probably mean, and that ignoring them falls through to the expensive lookup.

jakebeal (approved):

PR #118, opened 2022-02-24
Add a reference cache context manager

tcmitchell:
Add a context manager for a reference cache so that it can be used in a `with` statement. The resulting cache can be passed explicitly to find_child or find_top_level, or it can be passed implicitly because the cache is stored on a hidden attribute on the document.

Sample usage, passing the cache explicitly:
```python
with cached_references(doc) as cache:
        find_top_level(component1.sequences[0], cache)
```

Sample usage, passing the cache implicitly:
```python
with cached_references(doc):
        find_top_level(component1.sequences[0])
```

### DIFF
--- a/sbol_utilities/calculate_sequences.py
+++ b/sbol_utilities/calculate_sequences.py
@@ -4,7 +4,8 @@ from typing import List, Tuple, Union, Iterable
 
 import sbol3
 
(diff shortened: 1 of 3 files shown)


jakebeal (approved):
This is great --- I'm ready to start using it.

PR #120, opened 2022-02-24
Contained components & outgoing links

jakebeal:
Improve `component.contained_components` function
Also add an outgoing_links function that will be used for finding objects that need to be pulled into a document.


sbol_utilities/helper_functions.py
```diff
@@ -277,6 +282,78 @@ def has_plasmid_role(x):
...
+                _ = find_top_level(r)
+            except TopLevelNotFound:
+                outgoing.add(str(r))
+            except ValueError:
+                pass  # ignore references to child objects
+
+    outgoing = set()
+    with cached_references(doc):
+        doc.traverse(collector)
+    return outgoing
```
tcmitchell, line 359: Suggestion: add a newline at the end of the file
```suggestion
    return outgoing

```

sbol_utilities/helper_functions.py
```diff
@@ -277,6 +282,78 @@ def has_plasmid_role(x):
         return sbol3.SO_CIRCULAR in obj.types
     elif isinstance(obj, sbol3.SubComponent):  # if it's a subcomponent, check its definition
         return is_plasmid(find_top_level(obj.instance_of))
-        #return is_plasmid(obj.instance_of.lookup())
     else:
         return False
+
+
+class SBOL3PassiveVisitor:
```
tcmitchell, line 289: I like this base class a lot. It makes using the visitor pattern much easier.

sbol_utilities/component.py
```diff
@@ -1,44 +1,78 @@
...
+                    find_top_level(sc.instance_of).accept(self)
+
+        def visit_implementation(self, i: sbol3.Implementation):
+            if not self.already_visited(i):
+                if i.built:
+                    find_top_level(i.built).accept(self)
+
+    visitor = ContainmentVisitor()
+    root_list = list(roots)
+    if root_list:  # can't build the cache unless there's at least component to walk
```
tcmitchell, line 71: Add "one" in comment
```suggestion
    if root_list:  # can't build the cache unless there's at least one component to walk
```

sbol_utilities/component.py
```diff
@@ -1,44 +1,78 @@
...
 
-from sbol_utilities.helper_functions import id_sort, find_child
+from sbol_utilities.helper_functions import id_sort, find_child, find_top_level, SBOL3PassiveVisitor, cached_references
 from sbol_utilities.workarounds import get_parent
 
 
 # TODO: consider allowing return of LocalSubComponent and ExternallyDefined
-def contained_components(roots: Union[sbol3.TopLevel, Iterable[sbol3.TopLevel]]) -> Set[sbol3.Component]:
+def contained_components(roots: Union[sbol3.TopLevel, Iterable[sbol3.TopLevel]]) \
+        -> set[sbol3.Component]:
```
tcmitchell, line 14: If it were me I would break the line before the closing paren to avoid the backslash:

```suggestion
def contained_components(roots: Union[sbol3.TopLevel, Iterable[sbol3.TopLevel]]
        ) -> set[sbol3.Component]:
```

That might also need some reindenting, not sure.

tcmitchell (approved):
Looks good. I made a few minor suggestions which you can ignore or incorporate at your discretion.

sbol_utilities/helper_functions.py
jakebeal: Do you want to import it up into pySBOL3, or leave it here in SBOL-utilities?

sbol_utilities/component.py
jakebeal: I realized this line didn't actually need a wrap with the current style limit

sbol_utilities/helper_functions.py
jakebeal: Fixed.

sbol_utilities/helper_functions.py
tcmitchell: Leave it here for now. I think we'll want to do a review in the future and decide what should get moved to pySBOL3. There are a lot of good nuggets here.

PR #135, opened 2022-06-23
add: genbank conversion support for multiple records and features

mohitdmak:
## Changes:
* Further added support for the new converter to parse files with multiple records and features, like `test/test_files/iGEM_SBOL2_imports.gb`.
* Added a unit test for the same, even though it's an identical function to the previous unit test, currently I think we can just keep it instead of calling the previous unit test function so that we can add a few more checks later if required as the complexity of GenBank files increases.
* Copied mappings for GenBank to SO ontologies mappings and vice versa from [here](https://github.com/SynBioDex/libSBOLj/blob/d99912e57e2d734d86576f6cdb6b24ce88cd6583/core2/src/main/java/org/sbolstandard/core2/GenBank.java#L143) into files `GB2SO`, `SO2GB` (modified the content via vim macros, makes it easier to parse them to create a CSV).
* The file `script.py` at the root level parses the above 2 files to create CSVs `gb2so.csv` and `so2gb.csv`. The script file, along with the intermediate GB2SO and SO2GB is temporarily needed, while we figure out what to do with "many to one" mappings. These can be deleted before finally merging.
* Added test file `test/test_files/iGEM_SBOL2_imports_from_genbank_to_sbol3_direct.nt`, generated from new converter; if the conversion code changes are confirmed, it can serve as a reference for the unit tests.

## TODO:
* ~Currently orientation of FeatureLocation is hardcoded to "inline orientation" ("https://identifiers.org/SO:0001030"); need to create a method to infer it from the GenBank file instead; not sure if it should be done here or in a different PR.~
* ~There should be defaults created for converting between GenBank and SO ontologies mappings if the key is not found in their respective dictionaries.~
* ~Figure out how to manage "many to one" mappings - whether to include all matched values as roles / only the first matched, etc.~ - New spreadsheet mappings [here](https://docs.google.com/spreadsheets/d/1X870i3NhO7xEhqhLXK4eravNd72x-O-xbrpmlT835nY) had no "many to one" mapping ambiguities.
* ~Also all FeatureLocation types are Range currently, need to infer whether to choose "Range" or "Cut" from the GenBank file.~
* ~Add CSVs as package data to store them along with the pip installation~
* ~Added logging, throw warnings for extra GenBank annotations not directly storable in sbol3 files~
* ~Error handling for ontology mappings CSV parsing, for users who may not have updated pip package~
* ~Removed temp scripts and files used to generate CSV~


tcmitchell:
I don't think this work is complete yet because the PR comment still says "TODO:". I'm going to leave some comments anyway.

1. `script.py` should not be at the top level, and needs a better name. Is this program intended to stick around for the long term, or was it intended to be run once to generate the CSV files? If long term, it should be put in a better place, and given a better name. If it was run once and that's it, it should be removed from git.
2. The files `GB2SO` and `SO2GB` have been added at the top level. How were these generated? Do we need to keep them around? If so, we should find a better place than the top level. We can discuss what a better place might be.
3. The files `gb2so.csv` and `so2gb.csv` are long term files, and as such should get moved to the `sbol_utilities` directory. They should be added to `setup.py` as data files so that they get installed when the package is installed. And the code in sbol3_genbank_conversion.py should access them by locating them in a relative path to `__file__`. (And this needs to be done in a platform independent way.) There are examples I can point you to. See `package_data` in `setup.py` and look around to see how `sbolgraph-standalone.js` is located by whatever python code executes it.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -1,48 +1,112 @@
...
+                                "test/test_files", "BBa_J23101.gb")
+SAMPLE_SBOL3_FILE_1 = os.path.join(os.path.abspath(os.path.curdir), "test/test_files",
+                                "BBa_J23101_from_genbank_to_sbol3_direct.nt")
+SAMPLE_GENBANK_FILE_2 = os.path.join(os.path.abspath(os.path.curdir), "test/test_files",
+                                "iGEM_SBOL2_imports.gb")
+SAMPLE_SBOL3_FILE_2 = os.path.join(os.path.abspath(os.path.curdir), "test/test_files",
+                                "iGEM_SBOL2_imports_from_genbank_to_sbol3_direct.nt")
+GB2SO_MAPPINGS = {}
+SO2GB_MAPPINGS = {}
+SO_SEQ_FEAT_ROLE_NS = "http://identifiers.org/so/"
```
tcmitchell, line 28: Is this the right namespace? pySBOL3 uses `https://identifiers.org/SO:`.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -1,48 +1,112 @@
...
+class GenBank_SBOL3_Converter:
+    def create_GB2SO_role_mappings(self, gb2so_csv: str = GB2SO_MAPPINGS_CSV, so2gb_csv: str = SO2GB_MAPPINGS_CSV):
+        """Reads 2 CSV Files containing mappings for converting between GenBank and SO ontologies roles
+        :param gb2so_csv: path to read genbank to so conversion csv file
+        :param so2gb_csv: path to read so to genbank conversion csv file
+        """
+        with open(gb2so_csv, mode='r') as csv_file:
+            csv_reader = csv.DictReader(csv_file)
+            for row in csv_reader:
+                GB2SO_MAPPINGS[row["GenBank_Ontology"]] = row["SO_Ontology"]
```
tcmitchell, line 42: `GB2SO_MAPPINGS` should be an instance variable, like `self.gb2so_map`, not a global variable. The same holds for `SO2GB_MAPPINGS` below.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -1,48 +1,112 @@
...
+        with open(so2gb_csv, mode='r') as csv_file:
+            csv_reader = csv.DictReader(csv_file)
+            for row in csv_reader:
+                SO2GB_MAPPINGS[row["SO_Ontology"]] = row["GenBank_Ontology"]
+
+
+    def convert_genbank_to_sbol3(self, gb_file: str, sbol3_file: str = "sbol3.out", namespace: str =
+                                TEST_NAMESPACE, write: bool = False) -> sbol3.Document:
+        """Convert a GenBank document on disk into an SBOL3 document
```
tcmitchell, line 51: Great job with the docstring here!

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -1,48 +1,112 @@
...
+        :return: SBOL3 document containing converted materials
+        """
+        # create sbol3 document, and record parser handler for gb file
+        sbol3.set_namespace(namespace)
+        doc = sbol3.Document()
+        # access records by parsing gb file using SeqIO class
+        records = list(SeqIO.parse(gb_file, "genbank").records)
+        for record in records:
+            # NOTE: Currently we assume only linear or circular topology is possible
+            COMP_TYPES.append(sbol3.SO_LINEAR
```
tcmitchell, line 69: Will `COMP_TYPES` keep growing and growing in size with each record processed since it is a global variable?

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -1,48 +1,112 @@
...
+            doc.add(comp)
+            # NOTE: Currently we use a fixed method of encoding (IUPAC)
+            seq = sbol3.Sequence(identity=record.name + "_sequence",
+                                 elements=str(record.seq.lower()),
+                                 encoding=SEQUENCE_ENCODING)
+            doc.add(seq)
+            comp.sequences = [seq]
+            if record.features:
+                # create updated py dict to store mappings between gb and so ontologies
+                self.create_GB2SO_role_mappings(gb2so_csv=GB2SO_MAPPINGS_CSV, so2gb_csv=SO2GB_MAPPINGS_CSV)
```
tcmitchell, line 84: does this mean that both CSV files are read in for each record that gets processed? Is it necessary to read these files for every record processed?

tcmitchell:
I have made a few comments even though I don't think this PR is complete. I'm happy to walk through any of these comments with you if they're not clear. Thanks!

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Actually, I found the NS which I wrote in `test/test_files/iGEM_SBOL2_imports.nt` . But I see that both of our URLs do resolve (eg: https://identifiers.org/SO:0000167 and http://identifiers.org/so/SO:0000167). What should I finally keep here?

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Fixed this

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Thanks! I realized there would need to be many more helper methods like create_mappings, so making them into a class with docstrs would be clean.

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Yes, I agree it wouldn't be efficient to parse both CSVs for every record.
* I have fixed this to parse only once for a GenBank file.
* Also added a flag parameter to only let 1 of the 2 CSVs be parsed, based on our direction of conversion.

mohitdmak:
> I don't think this work is complete yet because the PR comment still says "TODO:". I'm going to leave some comments anyway.
My bad, the PR was mostly complete, I forgot to update the description; thanks for the comments though!

> 1. `script.py` should not be at the top level, and needs a better name. Is this program intended to stick around for the long term, or was it intended to be run once to generate the CSV files? If long term, it should be put in a better place, and given a better name. If it was run once and that's it, it should be removed from git.
I don't think it makes sense to have the script along with those 2 "GB2SO" and "SO2GB" files, as if there needs to be any change in mappings, later on, we can just edit the CSVs directly right?

> 2. The files `GB2SO` and `SO2GB` have been added at the top level. How were these generated? Do we need to keep them around? If so, we should find a better place than the top level. We can discuss what a better place might be.
Same as above, we should probably delete these before merging.

> 3. The files `gb2so.csv` and `so2gb.csv` are long term files, and as such should get moved to the `sbol_utilities` directory. They should be added to `setup.py` as data files so that they get installed when the package is installed. And the code in sbol3_genbank_conversion.py should access them by locating them in a relative path to `__file__`. (And this needs to be done in a platform independent way.) There are examples I can point you to. See `package_data` in `setup.py` and look around to see how `sbolgraph-standalone.js` is located by whatever python code executes it.
Oh, this makes sense, I will look into this next.

mohitdmak:
Hi, I have added checks for creating "Range"/"Cut" for FeatureLocation, and also for the orientation to use.
Though I am not sure whether the implementation is accurate, mostly because of the way BioPython parses GenBank files, much of the content from which we can recognize orientations (like the "complement" keyword in locations line, or "20^21" to denote cuts) gets converted to different fields provided by BioPython.
* I have noticed the location's `strand` value changes from 1 to -1 on using the "complement" keyword
* Also, the start and end positions are equal only when "^" is used, neither "276", not "276..276" would give equal start/end positions.

Is the approach valid?

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -1,48 +1,140 @@
...
+        :param write: writes the generated sbol3 document in SORTED_NTRIPLES
+                          format to provided sbol3_file path
+        :return: SBOL3 document containing converted materials
+        """
+        # create sbol3 document, and record parser handler for gb file
+        sbol3.set_namespace(namespace)
+        doc = sbol3.Document()
+        # access records by parsing gb file using SeqIO class
+        logging.info("Parsing Genbank records using SeqIO class.")
+        records = list(SeqIO.parse(gb_file, "genbank").records)
```
tcmitchell, line 77: There's probably no need to read the entire genbank file into memory like this. You can instead iterate over the records one at a time a few lines down with:

```python
    for record in SeqIO.parse(gb_file, "genbank"):
```

setup.py
```diff
@@ -60,6 +60,9 @@
                                 'sbol-diff=sbol_utilities.sbol_diff:main']
       },
       packages=['sbol_utilities'],
-      package_data={'sbol_utilities': ['sbolgraph-standalone.js']},
+      package_data={
+          'sbol_utilities': ['sbolgraph-standalone.js'],
+          'sbol_utilities': ['gb2so.csv', 'so2gb.csv']
+      },
```
tcmitchell, line 66: I don't think this will work because there are two entries in the `package_data` dict for `'sbol_utilities'`. These files should all go in one list, and preferably vertically spaced with one file name per line.
```suggestion
      package_data={
          'sbol_utilities': [
              'gb2so.csv',
              'sbolgraph-standalone.js',
              'so2gb.csv',
          ],
      },
```

test/test_files/BBa_J23101_from_genbank_to_sbol3_direct.nt
```diff
@@ -1,14 +1,14 @@
...
-<https://testing.sbol3.genbank/BBa_J23101> <http://sbols.org/v3#role> <https://identifiers.org/SO:0000804> .
-<https://testing.sbol3.genbank/BBa_J23101> <http://sbols.org/v3#type> <https://identifiers.org/SBO:0000251> .
-<https://testing.sbol3.genbank/BBa_J23101> <http://sbols.org/v3#type> <https://identifiers.org/SO:0000987> .
-<https://testing.sbol3.genbank/BBa_J23101> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://sbols.org/v3#Component> .
-<https://testing.sbol3.genbank/BBa_J23101_sequence> <http://sbols.org/v3#displayId> "BBa_J23101_sequence" .
-<https://testing.sbol3.genbank/BBa_J23101_sequence> <http://sbols.org/v3#elements> "tttacagctagctcagtcctaggtattatgctagc" .
-<https://testing.sbol3.genbank/BBa_J23101_sequence> <http://sbols.org/v3#encoding> <https://identifiers.org/edam:format_1207> .
-<https://testing.sbol3.genbank/BBa_J23101_sequence> <http://sbols.org/v3#hasNamespace> <https://testing.sbol3.genbank/> .
-<https://testing.sbol3.genbank/BBa_J23101_sequence> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://sbols.org/v3#Sequence> .
+<https://test.sbol3.genbank/BBa_J23101> <http://sbols.org/v3#description> "constitutive promoter family member" .
```
tcmitchell, line 2: Is this change in test URI intentional?

test/test_genbank_sbol3_direct.py
```diff
@@ -1,21 +1,39 @@
...
+from sbol_utilities.sbol3_genbank_conversion import GenBank_SBOL3_Converter, TEST_NAMESPACE
 
 class TestGenBank2SBOL3(unittest.TestCase):
+    # Create converter instance
+    converter = GenBank_SBOL3_Converter()
     def test_simple_file_1(self):
         """Test conversion of a simple genbank file with a single sequence"""
-        sbol3.set_namespace("https://testing.sbol3.genbank/")
-        TEST_OUTPUT_SBOL3 = SBOL3_FILE_1 + ".test" 
+        SAMPLE_GENBANK_FILE_1 = str(Path(__file__).parent) + "/test_files/" + "BBa_J23101.gb"
```
tcmitchell, line 12: This style of path appending won't work on all platforms. Please use the file append operations. If you are using `pathlib.Path` you can do it like this:
```python
Path(__file__).parent / 'test_files' / 'BBa_J23101.gb'
```
`pathlib.Path` defines the `/` operator to be the path append operation in a cross-platform way.

tcmitchell (changes requested):
In addition to responding to the comments, please remove all files that should not be part of the pull request.

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: I see, I have edited this for both records and features in the latest commit.

test/test_files/BBa_J23101_from_genbank_to_sbol3_direct.nt
mohitdmak: Oh, I had previously shortened it to make it visually easier to debug file diffs, I guess, we can let it be this way?

mohitdmak:
> In addition to responding to the comments, please remove all files that should not be part of the pull request.


Addressed the review comments; in addition:
* I have made the logger to throw out warnings upon encountering GenBank annotations not directly storable in sbol3 files.
* Added error handling for parsing ontology mappings, in case the user does not have the CSVs as data files in the package.
* Also formatted the code using [black](https://black.readthedocs.io/en/stable/), do let me know if it looks fine so that we can continue using it further.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -1,48 +1,221 @@
...
+                        )
+                    else:
+                        locs = sbol3.Range(
+                            sequence=seq,
+                            start=int(gb_loc.start),
+                            end=int(gb_loc.end),
+                            orientation=feat_orientation,
+                        )
+                    # Obtain sequence feature role from gb2so mappings
+                    feat_role = sbol3.SO_NS
```
tcmitchell, line 191: This doesn't feel right to me. Are you sure about this? This will add `'https://identifiers.org/SO:'` to the list of roles? Is that really what you want?

sbol_utilities/sbol3_genbank_conversion.py
tcmitchell: See below, I have asked a question about the use of `sbol3.SO_NS`. I'm closing this conversation as resolved since you fixed the thing that concerned me here.

jakebeal:
I'm not wild about the code style from `black` either, but as long as it's not changing pre-existing code I can live with it.

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: No this won't directly get added to feature's roles, as the next if/else would append the exact no after the "SO:" and then the final string would get appended to the feature's roles.
There was one issue though, I must use `sbol3.SO_NS[:-3]`, as we don't want "SO:" to get repeated twice. 
This produces proper feat roles in the test file.

tcmitchell (approved):
Thanks for responding to all the comments.

PR #136, opened 2022-06-30
add: new converter to existing console script with fallback option to legacy converter

mohitdmak:
## Changes:
* Existing executable commands `sbol2genbank`, `genbank2sbol`, `sbol-converter` now have an added flag `--force-new-converter`, which upon invoking would carry out the given conversion process using the new offline converter.
* This conversion would throw Errors in case of not being able to carry out certain parts of the conversion process.
* A new unit test, `test_genbank_conversion_new_converter`, has been added, which will test the converted output using the new force flag.
* This test uses the previously added test file: `BBa_J23101_from_genbank_to_sbol3_direct.nt` to convert the sbol3 document to genbank which is compared against the test file: `BBa_J23101_from_sbol3_direct.gb`.

## TODO:
* Add an option to revert to the legacy converter in case of any errors while using the new converter. (Long Term Issue)

### DIFF
--- a/sbol_utilities/conversion.py
+++ b/sbol_utilities/conversion.py
@@ -294,7 +294,7 @@ def convert_from_fasta(path: str, namespace: str, identity_map: Dict[str, str] =
 
 # TODO: Figure out how to support multiple namespaces like we do for FASTA: currently, importing from multiple
 #  namespaces will not work correctly
-def convert_from_genbank(path: str, namespace: str, allow_genbank_online: bool = False) -> sbol3.Document:
+def convert_from_genbank(path: str, namespace: str, allow_genbank_online: bool = False, force_new_converter: bool = False) -> sbol3.Document:
     """Convert a GenBank document on disk into an SBOL3 document
     Specifically, the GenBank document is first imported to SBOL2, then converted from SBOL2 to SBOL3
 
@@ -303,6 +303,18 @@ def convert_from_genbank(path: str, namespace: str, allow_genbank_online: bool =
     :param allow_genbank_online: Use the online converter, rather than the local converter
     :return: SBOL3 document containing converted materials
     """
+    if force_new_converter:
+        converter = GenBank_SBOL3_Converter()
+        doc = converter.convert_genbank_to_sbol3(gb_file=path, namespace=namespace, write=False)
+        return doc
+        # if not doc3:
+        #     logging.error(f"No SBOL3 output document was produced by new converter;\n \
+        #                   Please make sure you have the latest pip package installed.")
+        #     if allow_legacy_converter:
+        #         logging.info(f"Error encountered with conversion using new offline converter, reverting to legacy online converter.")
+        #         doc3 = convert_from_genbank(input_file, namespace, args_dict['allow_genbank_online'])
+        #     else:
+        #         return
     doc2 = sbol2.Document()
     sbol2.setHomespace(namespace)
     # Convert document offline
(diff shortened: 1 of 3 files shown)


sbol_utilities/conversion.py
```diff
@@ -395,7 +396,17 @@ def command_line_converter(args_dict: Dict[str, Any]):
...
-        doc3 = convert_from_genbank(input_file, namespace, args_dict['allow_genbank_online'])
+        allow_legacy_converter = args_dict['allow_legacy_converter']
+        converter = GenBank_SBOL3_Converter()
+        doc3 = converter.convert_genbank_to_sbol3(gb_file=input_file, sbol3_file=output_file, 
+                                                  namespace=namespace, write=False)
+        if not doc3:
+            logging.error(f"No SBOL3 output document was produced by new converter;\n    Please make sure you have the latest pip package installed.")
+            if allow_legacy_converter:
+                logging.info(f"Error encountered with conversion using new offline converter, reverting to legacy online converter.")
+                doc3 = convert_from_genbank(input_file, namespace, args_dict['allow_genbank_online'])
```
jakebeal, line 407: The switching logic should happen _inside_ this call, not here, in order to support library usage.

mohitdmak: I have edited this, also I needed to add a check if returning doc3 is none while outputting (writing) the doc3 in the `command_line_converter` method.

Also, how should I go about writing tests for this?
Should I add final outputs from my converter into `test/test_files` and write test cases where the `--force-new-converter` is used?

sbol_utilities/conversion.py
```diff
@@ -302,6 +303,18 @@ def convert_from_genbank(path: str, namespace: str, allow_genbank_online: bool =
     :param allow_genbank_online: Use the online converter, rather than the local converter
     :return: SBOL3 document containing converted materials
     """
+    if force_new_converter:
+        converter = GenBank_SBOL3_Converter()
+        doc = converter.convert_genbank_to_sbol3(gb_file=path, namespace=namespace, write=False)
+        return doc
+        # if not doc3:
```
jakebeal, line 310: Remove commented materials.

sbol_utilities/conversion.py
```diff
@@ -407,7 +423,7 @@ def command_line_converter(args_dict: Dict[str, Any]):
         raise ValueError(f'Unknown file type {input_file_type}; should have been caught earlier')
 
     # Convert output from SBOL3
-    logging.info('Writing output file '+output_file)
+    if doc3: logging.info('Writing output file '+output_file)
```
jakebeal, line 426: Remove this if, and the one below, since we shouldn't ever have a None returned.

sbol_utilities/conversion.py
```diff
@@ -474,6 +490,8 @@ def genbank2sbol():
                         help='Print running explanation of conversion process')
     parser.add_argument('--allow-genbank-online', dest='allow_genbank_online', action='store_true', default=False,
                         help='Perform GenBank conversion using online converter')
+    parser.add_argument('--force-new-converter', dest='force_new_converter', action='store_true', default=False,
```
jakebeal, line 493: You also need to add this argument for `def main()` as well

jakebeal (changes requested):

sbol_utilities/conversion.py
```diff
@@ -302,6 +303,18 @@ def convert_from_genbank(path: str, namespace: str, allow_genbank_online: bool =
     :param allow_genbank_online: Use the online converter, rather than the local converter
     :return: SBOL3 document containing converted materials
     """
+    if force_new_converter:
+        converter = GenBank_SBOL3_Converter()
+        doc = converter.convert_genbank_to_sbol3(gb_file=path, namespace=namespace, write=False)
+        return doc
```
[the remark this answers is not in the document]
mohitdmak: Done.

sbol_utilities/conversion.py
mohitdmak: Also added it to the `sbol2genbank()` function.

mohitdmak:
@jakebeal  Along with fixing the code reviews, I have added a unit test for the new conversion here.

sbol_utilities/conversion.py
mohitdmak: Yep, I have called the `convert_from_genbank()` directly in the unit test.

jakebeal (approved):

PR #143, opened 2022-07-14
add: new support for SBOL3 to GenBank conversion

mohitdmak:
## Changes:
* Newly added `convert_sbol3_to_genbank` method of the `GenBank_SBOL3_Converter` Class supports converting from sbol3 to Genbank file formats.
* Also, using the flag `--force-new-converter` through the `sbol2genbank` and `sbol-converter` will also now invoke the new method.
* Unit test `test_sbol3togb_1` has been added to test the conversion by directly invoking the new method.
* Unit test `test_genbank_conversion_new_converter` has been added to test the conversion by invoking it from the console script, using the previously added flag `--force-new-converter.`
* 2 New Test files: `BBa_J23101_from_sbol3_direct.gb` and `iGEM_SBOL2_imports_from_sbol3_direct.gb` has been added to aid the above 2 unit tests.

### Update:
1. Throws ValueError when the sbol3 component has more than 1 sequence. (TODO: Use more specific ErrorClass later)
2. The converter used to pick up features from a sbol3 Component in random order every time the conversion was called, leading to different file outputs each time. We discussed this in our last meeting, and I have implemented a lexicographic sorting of features of each component based on the feature location's start and endpoints. This ensures the output stays the same always for a specific input sbol3 document.
3. I have made the logger throw warnings to the console whenever, a feature translation lookup does not have any row in the CSV with the role in the given SBOL3/GenBank feature. The logger warns the user of this and mentions the default ontology term it will be using.

### Clarification Needed:
- [x]  Regarding update 1.; I have made the sequence object required while defining a BioPython `SeqRecord` class as `None` when a component does not have a sequence. Is this fine?
    * Moved to https://github.com/SynBioDex/SBOL-utilities/issues/147
- [ ] Regarding update 2.; There may be multiple feature locations for every feature; in this case, how should the sorting be implemented? Currently, it just sorts features based on the start/end points assuming there is only 1 location.
    * Moved to https://github.com/SynBioDex/SBOL-utilities/issues/148
- [x] Resolvement of [this issue](https://github.com/SynBioDex/SBOL-utilities/issues/146).
- [ ] We decided to store the `VERSION` annotation for GenBank records as "1" by default for now since the extra annotation fields are not currently stored in SBOL3 files during conversion from GenBank to SBOL3. However, this does not seem to reflect in converted GenBank files (the file shows no difference in the `VERSION` field, even after specifying it during the record's creation). 
    * I think this is covered by https://github.com/SynBioDex/SBOL-utilities/issues/150. If not, a new issue should be opened specifically for the VERSION field.

## TODO:

- [ ] A point was brought up in the last discussion to remove the `write` argument altogether in both the directions of conversion and have the logic implemented separately. We would need to discuss the details of its implementation.
- [ ] Also, I learned that features of type `SubComponents` in SBOL3 components would also need to be stored in the GenBank file. Currently, the converter ignores all features other than `SequenceFeature` while converting to GenBank. This would again need some discussion to implement.
- [ ] There can also be multiple feature locations, but it turns out storing multiple feature locations for a single GenBank feature is not that straightforward. SBOL3 allows for a simple way to store locations in an array, but doing something similar for GenBank will be required too.

### DIFF
--- a/sbol_utilities/conversion.py
+++ b/sbol_utilities/conversion.py
@@ -402,10 +402,7 @@ def command_line_converter(args_dict: Dict[str, Any]):
     if input_file_type == 'FASTA':
         doc3 = convert_from_fasta(input_file, namespace)
     elif input_file_type == 'GenBank':
-        doc3 = convert_from_genbank(path=input_file,
-                                    namespace=namespace,
-                                    allow_genbank_online=args_dict['allow_genbank_online'],
-                                    force_new_converter=args_dict['force_new_converter'])
+        doc3 = convert_from_genbank(input_file, namespace, args_dict['allow_genbank_online'], args_dict['force_new_converter'])
     elif input_file_type == 'SBOL2':
         doc2 = sbol2.Document()
         doc2.read(input_file)
@@ -421,7 +418,7 @@ def command_line_converter(args_dict: Dict[str, Any]):
     if output_file_type == 'FASTA':
         convert_to_fasta(doc3, output_file)
     elif output_file_type == 'GenBank':
-        convert_to_genbank(doc3=doc3, path=output_file, allow_genbank_online=args_dict['allow_genbank_online'], force_new_converter=args_dict['force_new_converter'])
+        convert_to_genbank(doc3, output_file, args_dict['allow_genbank_online'], args_dict['force_new_converter'])
     elif output_file_type == 'SBOL2':
         doc2 = convert3to2(doc3)
         validate_online = sbol2.Config.getOption(sbol2.ConfigOptions.VALIDATE_ONLINE)
--- a/sbol_utilities/conversion.py
+++ b/sbol_utilities/conversion.py
@@ -320,7 +320,7 @@ def convert_from_genbank(path: str, namespace: str, allow_genbank_online: bool =
     return doc
 
 
-def convert_to_genbank(doc3: sbol3.Document, path: str, allow_genbank_online: bool = False) \
+def convert_to_genbank(doc3: sbol3.Document, path: str, allow_genbank_online: bool = False, force_new_converter: bool = False) \
         -> List[SeqRecord.SeqRecord]:
     """Convert an SBOL3 document to a GenBank file, which is written to disk
     Note that for compatibility with version control software, if no prov:modified term is available on each Component,
@@ -331,6 +331,9 @@ def convert_to_genbank(doc3: sbol3.Document, path: str, allow_genbank_online: bo
     :param allow_genbank_online: use the online converter rather than the local converter
     :return: BioPython SeqRecord of the GenBank that was written
     """
+    if force_new_converter:
+        converter = GenBank_SBOL3_Converter()
+        return converter.convert_sbol3_to_genbank(sbol3_file=None, doc=doc3, gb_file=path, write=True)
     # first convert to SBOL2, then export to a temp GenBank file
     doc2 = convert3to2(doc3)
 
@@ -418,7 +421,7 @@ def command_line_converter(args_dict: Dict[str, Any]):
     if output_file_type == 'FASTA':
         convert_to_fasta(doc3, output_file)
     elif output_file_type == 'GenBank':
-        convert_to_genbank(doc3, output_file, args_dict['allow_genbank_online'])
+        convert_to_genbank(doc3=doc3, path=output_file, allow_genbank_online=args_dict['allow_genbank_online'], force_new_converter=args_dict['force_new_converter'])
     elif output_file_type == 'SBOL2':
         doc2 = convert3to2(doc3)
         validate_online = sbol2.Config.getOption(sbol2.ConfigOptions.VALIDATE_ONLINE)
--- a/sbol_utilities/sbol3_genbank_conversion.py
+++ b/sbol_utilities/sbol3_genbank_conversion.py
@@ -189,6 +189,7 @@ class GenBank_SBOL3_Converter:
                     if self.gb2so_map.get(gb_feat.type):
                         feat_role += self.gb2so_map[gb_feat.type]
                     else:
+                        logging.warning(f"Feature type: `{gb_feat.type}` for feature: `{gb_feat.qualifiers['label'][0]}` of record: `{record.name}` has no corresponding ontology term for SO, using the default SO term, {self.DEFAULT_SO_TERM}")
                         feat_role += self.DEFAULT_SO_TERM
                     feat = sbol3.SequenceFeature(
                         locations=[locs],
@@ -235,9 +236,13 @@ class GenBank_SBOL3_Converter:
         for obj in doc.objects:
             if isinstance(obj, sbol3.Component):
                 logging.info(f"Parsing component - `{obj.display_id}` in sbol3 document.")
-                # TODO: can component have multiple sequences? How should we handle Seq objects for those
-                obj_seq = doc.find(obj.sequences[0])
-                seq = Seq(obj_seq.elements.upper())
+                # NOTE: A single component/record cannot have multiple sequences
+                seq = None # If no sequence is found for a component
+                if obj.sequences and len(obj.sequences) == 1:
+                    obj_seq = doc.find(obj.sequences[0])
+                    seq = Seq(obj_seq.elements.upper())
+                elif len(obj.sequences) > 1:
+                    raise ValueError(f"Component `{obj.display_id}` of given SBOL3 document has more than 1 sequnces (`{len(obj.sequences)}`). This is invalid; a component may only have 1 or 0 sequences.")
                 # TODO: "Version" annotation information currently not stored when converted genbank to sbol3
                 seq_rec = SeqRecord(
                     seq=seq,
@@ -271,7 +276,7 @@ class GenBank_SBOL3_Converter:
                             if obj_feat_loc.orientation == sbol3.SO_REVERSE:
                                 feat_strand = self.BIO_STRAND_REVERSE
                             elif obj_feat_loc.orientation != sbol3.SO_FORWARD:
-                                raise ValueError(f"Location orientation: '{obj_feat_loc.orientation}' for feature: '{obj_feat.name}' of component: '{obj.display_id}' is not a valid orientation.\n Valid orientations are '{sbol3.SO_FORWARD}', '{sbol3.SO_REVERSE}'")
+                                raise ValueError(f"Location orientation: `{obj_feat_loc.orientation}` for feature: `{obj_feat.name}` of component: `{obj.display_id}` is not a valid orientation.\n Valid orientations are `{sbol3.SO_FORWARD}`, `{sbol3.SO_REVERSE}`")
                             # TODO: Raise custom converter class ERROR for `else:`
                             feat_loc = FeatureLocation(
                                 start=obj_feat_loc.start,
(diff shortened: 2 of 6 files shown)


sbol_utilities/conversion.py
```diff
@@ -418,7 +421,7 @@ def command_line_converter(args_dict: Dict[str, Any]):
     if output_file_type == 'FASTA':
         convert_to_fasta(doc3, output_file)
     elif output_file_type == 'GenBank':
-        convert_to_genbank(doc3, output_file, args_dict['allow_genbank_online'])
+        convert_to_genbank(doc3=doc3, path=output_file, allow_genbank_online=args_dict['allow_genbank_online'], force_new_converter=args_dict['force_new_converter'])
```
jakebeal, line 424: You don't need to turn these into keyword arguments: it's better to make minimal changes where possible.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -58,7 +62,7 @@ class GenBank_SBOL3_Converter:
     gb2so_map = {}
     so2gb_map = {}
     default_SO_ontology = "SO:0000110"
-    default_GB_ontology = ""  # TODO: Whats the default here?
+    default_GB_ontology = "todo"  # TODO: Whats the default here?
```
jakebeal, line 65: Right name for these is "term", not "ontology" - the ontology is the full collection of organized terms.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -58,7 +62,7 @@ class GenBank_SBOL3_Converter:
     gb2so_map = {}
     so2gb_map = {}
     default_SO_ontology = "SO:0000110"
```
jakebeal, line 64: Make an issue to consider whether this should actually be "SO:0000001" (disagreeing with libSBOLj)

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -206,15 +210,114 @@ def convert_genbank_to_sbol3(
             doc.write(fpath=sbol3_file, file_format=sbol3.SORTED_NTRIPLES)
         return doc
 
+    def convert_sbol3_to_genbank(
+        self,
```
jakebeal, line 214: Minor style point: this is not the usual indentation for long argument lists.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -206,15 +210,114 @@ def convert_genbank_to_sbol3(
...
+                    description=obj.description,
+                    name=obj.display_id,
+                )
+                # TODO: hardcoded molecule_type as DNA, derivation?
+                seq_rec.annotations["molecule_type"] = "DNA"
+                # TODO: hardcoded topology as linear, derivation?
+                seq_rec.annotations["topology"] = "linear"
+                if obj.features:
+                    # converting all sequence features
+                    for obj_feat in obj.features:
```
jakebeal, line 267: Add a filter for taking only the SequenceFeature features.
Add a TODO for SubComponent features in the future.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -206,15 +210,114 @@ def convert_genbank_to_sbol3(
...
+                        logging.info(
+                            f"Parsing feature `{obj_feat.name}` for component `{obj.display_id}`"
+                        )
+                        obj_feat_loc = obj_feat.locations[0]
```
jakebeal, line 271: Todo for multiple locations

jakebeal (changes requested):
Just minor changes needed

tcmitchell (changes requested):
Overall this looks good. As with @jakebeal comments, just a few things to change.

sbol_utilities/conversion.py
mohitdmak: Fixed

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Oh, alright, I changed them.

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: I have edited this in the latest commit, but am not sure if even it is the correct way.
Maybe we should decide on a py formatter to handle these style issues?

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Added, we are just taking note of this currently right?

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -206,15 +210,114 @@ def convert_genbank_to_sbol3(
...
+                        feat_loc = FeatureLocation(
+                            start=obj_feat_loc.start - 1,
+                            end=obj_feat_loc.end,
+                            strand=feat_strand,
+                        )
+                        # TODO: Raise custom converter class ERROR for `else:`
+                        # FIXME: order of features not same as original genbank doc?
+                        obj_feat_role = obj_feat.roles[0]
+                        obj_feat_role = obj_feat_role[
+                            obj_feat.roles[0].index(":", 5) - 2 :
```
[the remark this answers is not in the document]
mohitdmak: * My bad, removed the `obj_feat.roles[0]` there.
* Actually, the `so2gb.csv` data file has rows of format `SO:xxxxxxx,<GenBank_Term>`, and the `obj_feat_role` returns the URI (i.e `https://identifiers.org/SO:xxxxxx`). So the `.index(":",5) - 2` would be the position of index from where the URI is `SO:xxxxxxx`.
* I have now added this in a comment.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -206,15 +210,114 @@ def convert_genbank_to_sbol3(
...
+
+        :param sbol3_file: path to read SBOL3 file from
+        :param gb_file: path to write GenBank file to, if write set to true
+        :param write: writes the generated genbank document to provided path
+        :return: Array of SeqRecord objects which comprise the generated GenBank document
+        """
+        if not doc:
+            doc = sbol3.Document()
+            doc.read(sbol3_file)
+        SEQ_RECORDS = []
```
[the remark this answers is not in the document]
mohitdmak: Okay, fixed

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -206,15 +210,114 @@ def convert_genbank_to_sbol3(
...
+            # TODO: Need better SBOL3-GenBank specific error classes in future
+            raise ValueError(
+                f"Required CSV data files are not present in your package.\n    Please reinstall the sbol_utilities package.\n \
+                Stopping current conversion process.\n    Reverting to legacy converter if new Conversion process is not forced."
+            )
+        # consider sbol3 objects which are components
+        logging.info(f"Parsing SBOL3 Document components using SBOL3 Document: \n{doc}")
+        for obj in doc.objects:
+            if isinstance(obj, sbol3.Component):
+                SEQ_REC_FEATURES = []
```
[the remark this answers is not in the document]
mohitdmak: Okay, fixed

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -206,15 +210,114 @@ def convert_genbank_to_sbol3(
...
+                        logging.info(
+                            f"Parsing feature `{obj_feat.name}` for component `{obj.display_id}`"
+                        )
+                        obj_feat_loc = obj_feat.locations[0]
+                        feat_strand = 1
+                        # feature strand value which denotes orientation of the location of the feature
+                        if obj_feat_loc.orientation == sbol3.SO_FORWARD:
+                            feat_strand = 1
+                        elif obj_feat_loc.orientation == sbol3.SO_REVERSE:
+                            feat_strand = -1
```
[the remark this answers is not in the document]
mohitdmak: * Alright, I have added these constants with comments and removed the redundant line checking for `SO_FORWARD`.
* Also, @jakebeal and I discussed in last week's meeting to raise a ValueError (ValueError for now, a custom ConverterClass Error later on) when the sbol3 feature's location's orientation is none of `SO_REVERSE` or `SO_FORWARD.`
* I have added a check for this to raise a ValueError for now, is this fine?

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Done, created [here](https://github.com/SynBioDex/SBOL-utilities/issues/146).

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -57,16 +61,14 @@
 class GenBank_SBOL3_Converter:
     gb2so_map = {}
     so2gb_map = {}
-    default_SO_ontology = "SO:0000110"
-    default_GB_ontology = ""  # TODO: Whats the default here?
+    DEFAULT_SO_TERM = "SO:0000110"
+    DEFAULT_GB_TERM = "todo"  # TODO: Whats the default here?
```
jakebeal, line 65:
```suggestion
    DEFAULT_GB_TERM = "misc_feature"
```
Per a conversation with @cjmyers today

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -206,15 +204,129 @@ def convert_genbank_to_sbol3(
...
+                if obj.features:
+                    # converting all sequence features
+                    for obj_feat in obj.features:
+                        # TODO: Also add ability to parse subcomponent feature type
+                        # Note: Currently we only parse sequence features from sbol3 to genbank
+                        if isinstance(obj_feat, sbol3.SequenceFeature):
+                            logging.info(
+                                f"Parsing feature `{obj_feat.name}` for component `{obj.display_id}`"
+                            )
+                            # TODO: There may be multiple locations for a feature from sbol3;
```
mohitdmak, line 270: @jakebeal, the TODO is present here for multiple feature locations

jakebeal: Huh... for some reason I was seeing only the old version. Thank you!

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
tcmitchell: Looks good, thanks.

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
tcmitchell: Yes, this is fine. Thanks!

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
tcmitchell: Thanks for adding the explanatory comment

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
tcmitchell: 👍

sbol_utilities/sbol3_genbank_conversion.py
tcmitchell: @jakebeal I see a TODO in the updated source file.

tcmitchell (approved):
Thanks for making the requested changes. I am approving, and we want to be sure @jakebeal also approves before merging.

We also need issues opened for the outstanding items in the PR description.

jakebeal (approved):

PR #157, opened 2022-07-25
digestion

Gonza10V:
digestion function


sbol_utilities/component.py
```diff
@@ -621,4 +622,84 @@ def part_in_backbone(identity: str, part: sbol3.Component, backbone: sbol3.Compo
     part_in_backbone_component.features.append(backbone_subcomponent)
     # adding topology
     part_in_backbone_component.types.append(topology_type)
-    return part_in_backbone_component, part_in_backbone_seq
\ No newline at end of file
+    return part_in_backbone_component, part_in_backbone_seq
+
+def digestion(reactant:sbol3.Component, restriction_enzymes:List[sbol3.ExternallyDefined], assembly_plan:sbol3.Component)-> Tuple[sbol3.Component, sbol3.Sequence]:
+    """Digests a Component using the provided restriction enzymes and creates a product Component and a digestion Interaction.
+
```
jakebeal, line 629: Note that other components may be produced too (e.g., a backbone that is being discarded), but that we are only representing the intended product.

sbol_utilities/component.py
```diff
@@ -621,4 +622,84 @@ def part_in_backbone(identity: str, part: sbol3.Component, backbone: sbol3.Compo
...
+    :param reactant: DNA to be digested as SBOL Component. 
+    :param restriction_enzymes: Restriction enzymes used  Externally Defined.
+    :return: A tuple of Component and Interaction.
+    """
+    if all(t != sbol3.SBO_DNA for t in reactant.types):
```
jakebeal, line 634: Should be equivalent to simpler: `sbol3.SBO_DNA not in reactant.types`

sbol_utilities/component.py
```diff
@@ -621,4 +622,84 @@ def part_in_backbone(identity: str, part: sbol3.Component, backbone: sbol3.Compo
...
+        raise TypeError(f'The reactant should has a DNA type. Types founded {reactant.types}.')
+    if len(reactant.sequences)!=1:
+        raise ValueError(f'The reactant needs to have only one sequence. The input reactant has {len(reactant.sequences)} sequences')
```
jakebeal, line 637:
```suggestion
        raise ValueError(f'The reactant needs to have precisely one sequence. The input reactant has {len(reactant.sequences)} sequences')
```

sbol_utilities/component.py
```diff
@@ -621,4 +622,84 @@ def part_in_backbone(identity: str, part: sbol3.Component, backbone: sbol3.Compo
...
+    for re in restriction_enzymes:
+        enzyme = Restriction.__dict__[re.name]
+        restriction_enzymes_pydna.append(enzyme)
+        assembly_plan.features.append(re)
+        modifier_participation = sbol3.Participation(roles=[sbol3.SBO_MODIFIER], participant=re)
+        participations.append(modifier_participation)
+
+    # Inform topology to PyDNA, if not found assuming circular. 
+    if is_circular(reactant):
+        circular=True
```
jakebeal, line 649: Shouldn't need to track both circular and linear, because linear = not circular

sbol_utilities/component.py
```diff
@@ -621,4 +622,84 @@ def part_in_backbone(identity: str, part: sbol3.Component, backbone: sbol3.Compo
...
+        linear=False
+    reactant_seq = reactant.sequences[0].lookup().elements
+    # Dseqrecord is from PyDNA package with reactant sequence
+    ds_reactant = Dseqrecord(reactant_seq, linear=linear, circular=circular)
+    digested_reactant = ds_reactant.cut(restriction_enzymes_pydna)
+
+    if len(digested_reactant)==0 or len(digested_reactant)>3:
+        raise NotImplementedError(f'Not supported number of products. Found{len(digested_reactant)}')
+    elif circular and len(digested_reactant)==2:
+        digested_reactant = ds_reactant.cut(restriction_enzymes_pydna)
```
jakebeal, line 665: Delete this; use line 660. Same for below in 668.

sbol_utilities/component.py
```diff
@@ -621,4 +622,84 @@ def part_in_backbone(identity: str, part: sbol3.Component, backbone: sbol3.Compo
...
+        linear=True
+    else: 
+        circular=True
+        linear=False
+    reactant_seq = reactant.sequences[0].lookup().elements
+    # Dseqrecord is from PyDNA package with reactant sequence
+    ds_reactant = Dseqrecord(reactant_seq, linear=linear, circular=circular)
+    digested_reactant = ds_reactant.cut(restriction_enzymes_pydna)
+
+    if len(digested_reactant)==0 or len(digested_reactant)>3:
```
jakebeal, line 662:
```suggestion
    if len(digested_reactant)<2 or len(digested_reactant)>3:
```

sbol_utilities/component.py
```diff
@@ -621,4 +622,84 @@ def part_in_backbone(identity: str, part: sbol3.Component, backbone: sbol3.Compo
...
+        raise NotImplementedError(f'Not supported number of products. Found{len(digested_reactant)}')
+    elif circular and len(digested_reactant)==2:
+        digested_reactant = ds_reactant.cut(restriction_enzymes_pydna) 
+        part_extract, backbone = digested_reactant
```
jakebeal, line 666: Need to figure out which one is the backbone.
Initial version: assume larger one is backbone.
Add a TODO to have it be based on content rather than size.

sbol_utilities/component.py
```diff
@@ -621,4 +622,84 @@ def part_in_backbone(identity: str, part: sbol3.Component, backbone: sbol3.Compo
...
+    elif linear and len(digested_reactant)==3:
+        digested_reactant = ds_reactant.cut(restriction_enzymes_pydna)
+        # check digested_reactant
+        prefix, part_extract, suffix = digested_reactant
+    else: raise NotImplementedError('The reactant has no valid topology type')
+    # Extracting roles from features
+    reactant_features_roles = []
+    for f in reactant.features:
```
jakebeal, line 674: I'm pretty sure this is incorrect. You want to inspect based on locations. Let's see what happens when you set up the test...

sbol_utilities/component.py
Gonza10V: here I would see if the topology is linear and if not  especified as linear, assume circular given that the other option is circular and if not informed it assumes linear. This can change if more DNA topologies are added to SBOL in future but is very unlikely in the short term. So what do you think about assesing linearity and if not assume that is circular?

sbol_utilities/component.py
Gonza10V: deleted extra cuts

sbol_utilities/component.py
jakebeal: The specification currently says that if topology is not circular, then one can assume linear.
I would thus recommend using that.

Gonza10V: now assuming linear if not circular

sbol_utilities/component.py
Gonza10V: implemented

sbol_utilities/component.py
Gonza10V: I added 'The product Component is assumed to be the insert for parts in backbone and the backbone for backbones.' to the docstring, that is enough to inform the user or should I specify what I look to define if it is a backbone or a part_in_backbone or that should be an unser input?

sbol_utilities/component.py
Gonza10V: implemented using `part_extract, backbone = sorted(digested_reactant, key=len)`

sbol_utilities/component.py
Gonza10V: docstring extended

sbol_utilities/component.py
Gonza10V: test passed

jakebeal (changes requested):
In addition to the two expected failures due to #231, there is a failing round-trip test. Can you please identify and resolve the issue?

Issue #150, opened 2022-07-21
Handle GenBank header lines in genbank converter


tcmitchell:
Genbank files have a number of header lines. The online sbol2-genbank converter used a custom namespace for these in order to round trip the data. Try the same approach or a similar approach. It would be nice if the approach was compatible with the sbol2-genbank converter.

PR #158, opened 2022-07-26
feat: storing and retrieving extraneous genbank properties in sbol3 components

mohitdmak:
Apologies for the size of this PR, but much of the diff is due to test files, and this whole PR feature had to be done in one go too.
This solves #150 

### Refactoring:
* All test files used exclusively by the unit tests for the new converter are moved to a subdirectory  `sbol3_genbank_conversion` inside `test/test_files.` 
* Also, a new file, `test_extra_properties.gb` has been added to test whether the extra annotations and properties of GenBank files persist during the conversion of GenBank -> SBOL3 -> GenBank. This file was taken up from the [SBOLTestSuite repository](https://github.com/SynBioDex/SBOLTestSuite) (`sequence2.gb`) and was chosen for this test because much of the annotations and properties in it are not the "standard/default" ones assigned by GenBank in the absence of them being explicitly provided.
* All the unit tests for the converter were using redundant code, so the following helper functions were created, and the unit tests  simply provided file paths as arguments to these:
    1. `_test_genbank_to_sbol3(self, SAMPLE_SBOL3_FILE: Path, SAMPLE_GENBANK_FILE: Path)` - To test conversion of test files from GenBank to SBOL3.
    2. `_test_sbol3_to_genbank(self, SAMPLE_SBOL3_FILE: Path, SAMPLE_GENBANK_FILE: Path)` - To test conversion of test files from SBOL3 to GenBank.
    3. `_test_round_trip_genbank(self, SAMPLE_GENBANK_FILE: Path)` - To test whether a GenBank file during a round trip to and from SBOL3 remains the same.
   
### Storing GenBank extra properties:
* The single source of reference for the attempt to store the extra information in SBOL3 as properties has been [these pysbol3 docs.](https://pysbol3.readthedocs.io/en/stable/extensions.html#defining-extension-properties)
* I created relevant properties for each extra information type *(like Date, File division, Keywords, Molecule type, Taxonomy, Organism, Source, etc.)*. But I realized we had to declare these property variables each time we wished to store them or retrieve them from an SBOL3 File as:
```
>>> c1 = sbol3.Component('http://example.org/sbol3/c1', sbol3.SBO_DNA)
>>> c1.x_coordinate = sbol3.IntProperty(c1, 'http://example.org/my_vis#x_coordinate', 0, 1)
```
* As an alternative to this, I created an inner class `Component_GenBank_Extension` to the Converter class, which instantiates these SBOL3 properties and assigns them to variables attached to a component object. To use this class *(which extends the default `sbol3.Component` class)*, I have created a builder method `build_component_genbank_extension` and have registered it as the builder method for Component objects in the Converter class's `__init__()` method.
* The entire process above is followed as explained [here](https://pysbol3.readthedocs.io/en/stable/extensions.html#example-1-override-a-core-class).
* The above method works fine for every type of extra information - except **References**. (The legacy converter does not bother to store **References** from a GenBank file as SBOL3 properties, but I think we should have it stored in the new converter for the sake of completeness. I would like to get your input on whether we should even attempt to store **References**).

### Issue with References:
* There can be any number of references for a GenBank record, and each of them has multiple **TextFields** *(like Authors, Title, Journal, Consortium, Date, etc.)*. Apart from this, each Reference also has a **location** field, which is an array of **FeatureLocations**.
* Currently, I have stored all the **TextFields** into a `TextListProperty` for each of them; and assigned them to a reference one by one while converting an SBOL3 file to GenBank.
* This could *kind of* be done for the **location** field too *(by using `OwnedObjectListProperty` with a **type constraint** of `FeatureLocation`), but would be a very messy way to do it. Since, while parsing SBOL3 -> GenBank, we wouldn't know how many locations to store in each **Reference**, we would also need to store **indexes** of end locations of each **Reference** in an array property for this to work.
* To avoid this, I had tried to create a custom **Property**  *(see the block of commented lines below the `Component_GenBank_Extension` class)*, but faced issues as apparently I need to return only **RDF Terms** from the **Property Mixin Class**, but they support only **Literal** Values. So this attempt led nowhere, but I have kept the commented lines for you to see if we can make this method work in some way.

## Updates:
1. As per discussions in the PR reviews, further development for the converter along these lines is to be made on separate PRs, the issues for which are filed here - 
2. Logging configurations and the `main()` method are removed from the main converter file, as we wish the logging configurations set at the place from where our converter is invoked to be used for its usage there.
3. Many of the previous unit test files have been edited to handle failures caused by diff in generated files (with extraneous properties stored) and stored reference test files (which did not have the properties previously).
4. Storage for new properties of **GenBank Identifiers (GI), Accessions, Sequence Version, Keywords**, etc. has been added to the `Component_GenBank_Extension` class.
5. The converter has been made to throw Errors now when it encounters "annotations" or "properties" in the GenBank file apart from those (standard) we already parse.

## TODO: 
- [ ] Currently, I have no idea how to solve the issue of storing the **FeatureLocations** of **References** from a GenBank file as SBOL3 properties.~*(That is if we are to proceed with storing them at all)*~.
    - **Note:** This is to be solved in a separate PR using the approach discussed in the below reviews; the related issue is filed [here.](https://github.com/SynBioDex/SBOL-utilities/issues/169)
- [ ] There needs to be some modifications made to the way the new `Component_GenBank_Extension` class is built on top of the `sbol3.Component` class as discussed in the below reviews; [here](https://github.com/SynBioDex/SBOL-utilities/issues/170) is the related issue.
- [ ] Regarding *Updates* point 5, we intend to provide users the option to either allow storing all "annotations" or "properties" that the converter comes across while parsing the GenBank file or only store the standard ones. An issue is created regarding the same [here.](https://github.com/SynBioDex/SBOL-utilities/issues/171)
- [x] Also, I am unsure if the class extension method for storing SBOL3 properties is the best way forward here.


sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -67,8 +70,104 @@ class GenBank_SBOL3_Converter:
...
+            """
+            # `types` is required and not known at build time.
+            # Supply a missing value to the constructor, then clear
+            # the missing value before returning the built object.
+            obj = self.Component_GenBank_Extension(identity=identity, types=[sbol3.PYSBOL3_MISSING], type_uri=type_uri)
+            # Remove the placeholder value
+            obj.clear_property(sbol3.SBOL_TYPE)
+            return obj
+        # set up logging
+        log_level = logging.INFO
```
jakebeal, line 88: I don't think that you want to set up logging here. Is there some reason you think it should be here?

jakebeal (changes requested):

jakebeal:
On the question of references: can you make extensions of CustomIdentified for Reference and for ReferenceLocation? They are objects, and that will let you deal with them as objects.

tcmitchell:
> On the question of references: can you make extensions of CustomIdentified for Reference and for ReferenceLocation? They are objects, and that will let you deal with them as objects.

I think the question of references should be deferred to a different issue and a different pull request. We can acknowledge that those are not handled yet, and allow this PR to be feature complete. There is a lot of good commentary in this PR's description that should form the basis of the new issue for handling GenBank References. And Jake's comment, quoted in this comment, should also be included. That work can be prioritized and undertaken in concert with other GenBank conversion tasks.

tcmitchell:
# Component_GenBank_Extension

* I think a better name is `GenbankComponent`. Python class names do not typically have underscores.
* I do not think a handler should be registered for this type.
  * As currently implemented, anyone who loads a file after the handler has been installed will get a `Component_GenBank_Extension` instead of a `sbol3.Component`. That might not be what they want or expect.
  * I think the GenBank converter and simply construct a `GenbankComponent` instead of a `Component` to get the extra properties. pySBOL3 will properly round trip the extra properties. We'll have to devise a way to detect the extra properties, which I'm sure can be done. We may want to manually add these properties to `Component` instances on the SBOL->GenBank translation so that we can more easily detect the properties. Or construction a `GenbankComponent` from each `Component` that we will translate from SBOL3 to GenBank.

As above, I think we should create a separate GitHub issue for this work to improve the converter. We can leave the existing code as it is.

test/test_genbank_sbol3_direct.py
```diff
@@ -15,96 +15,133 @@ class TestGenBankSBOL3(unittest.TestCase):
...
-        """Test conversion of a simple genbank file with a single sequence"""
-        SAMPLE_GENBANK_FILE_1 = Path(__file__).parent / "test_files" / "BBa_J23101.gb"
-        SAMPLE_SBOL3_FILE_1 = (
-            Path(__file__).parent
-            / "test_files"
-            / "BBa_J23101_from_genbank_to_sbol3_direct.nt"
-        )
-        sbol3.set_namespace(TEST_NAMESPACE)
-        TEST_OUTPUT_SBOL3 = str(SAMPLE_SBOL3_FILE_1) + ".test"
+    def _test_genbank_to_sbol3(self, SAMPLE_SBOL3_FILE: Path, SAMPLE_GENBANK_FILE: Path):
```
tcmitchell, line 18: I'm glad to see a helper function here. This is a good software design choice.

tcmitchell, line 18: Is there a reason for the arguments of this method to be in upper case? That is unusual. If there is a reason, please document it. If there is not a reason, please switch to lower case.

test/test_genbank_sbol3_direct.py
```diff
@@ -15,96 +15,133 @@ class TestGenBankSBOL3(unittest.TestCase):
...
+        """Helper method to test conversion of a given GenBank file to SBOL3 using new converter.
+        :param SAMPLE_SBOL3_FILE: Path of expected SBOL3 converted file
+        :param SAMPLE_GENBANK_FILE: Path of given GenBank file to convert
+        """
+        TEST_OUTPUT_SBOL3 = str(SAMPLE_SBOL3_FILE) + ".test"
```
tcmitchell, line 23: Is there a reason for this variable to be in upper case? It is unusual for a local variable to be upper case. If there is a reason, please document it. If there is not a reason, please switch to lower case.

test/test_genbank_sbol3_direct.py
```diff
@@ -15,96 +15,133 @@ class TestGenBankSBOL3(unittest.TestCase):
...
         sbol3_file_1.read(
-            location=str(SAMPLE_SBOL3_FILE_1), file_format=sbol3.SORTED_NTRIPLES
+            location=str(SAMPLE_SBOL3_FILE), file_format=sbol3.SORTED_NTRIPLES
         )
         assert not doc_diff(
             test_output_sbol3, sbol3_file_1
-        ), f"Converted SBOL3 file: {TEST_OUTPUT_SBOL3} not identical to expected file: {SAMPLE_SBOL3_FILE_1}"
+        ), f"Converted SBOL3 file: {TEST_OUTPUT_SBOL3} not identical to expected file: {SAMPLE_SBOL3_FILE}"
+    
+    def _test_sbol3_to_genbank(self, SAMPLE_SBOL3_FILE: Path, SAMPLE_GENBANK_FILE: Path):
```
tcmitchell, line 39: Same as above regarding upper case or lower case. If the arguments are intentionally upper case a comment should explain why. Otherwise these should be lower case.

tcmitchell (changes requested):
Requested changes:
* The unit tests are failing
* Remove extraneous top level files `genbank.out`, `sbol.nt`, and `test.gb.
* There are a lot of upper case local variables and method arguments in `test_genbank_sbol3_direct.py`. 

Comment, not to be fixed:
* `convert_genbank_to_sbol3` and `convert_sbol3_to_genbank` are too long, and should be broken apart into a set of helper methods in a future pull request. Perhaps this should be a new GitHub issue for refactoring these methods.

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Actually, previously it was being set up inside the `main()` function, and the only thing it did was create an instance of the `GenBank_SBOL3_Converter` class. I used this for debugging and had logging setup here.

If our `GenBank_SBOL3_Converter` class is being invoked somewhere else in the library, we would intend to use the logging configs already defined in that file right? So should I move the logging setup back to `main()`?

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -320,13 +505,11 @@ def convert_sbol3_to_genbank(self, sbol3_file: str, doc: sbol3.Document = None,
...
 # Currently we don't parse input for gb and sbol3 files (hardcoded)
 def main():
-    log_level = logging.INFO
-    logging.getLogger().setLevel(level=log_level)
     converter = GenBank_SBOL3_Converter()
-    # converter.convert_genbank_to_sbol3(
-    #     gb_file=SAMPLE_GENBANK_FILE_2, sbol3_file=SAMPLE_SBOL3_FILE_2, write=True
-    # )
-    converter.convert_sbol3_to_genbank(sbol3_file="iGEM_SBOL2_imports_from_genbank_to_sbol3_direct.nt", write=True)
+    converter.convert_genbank_to_sbol3(
```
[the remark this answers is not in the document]
mohitdmak: Yes, these were simply used for debugging, will remove them before this PR is ready to merge.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -67,8 +70,104 @@ class GenBank_SBOL3_Converter:
...
+        """Overrides the sbol3 Component class to include fields to directly read and write 
+        extraneous properties of GenBank not storeable in any SBOL3 datafield.
+        :extends: sbol3.Component class
+        """
+        GENBANK_EXTRA_PROPERTY_NS = "http://www.ncbi.nlm.nih.gov/genbank"
+        def __init__(self, identity: str, types: Optional[Union[str, Sequence[str]]], **kwargs) -> None:
+            # instantiating sbol3 component object
+            super().__init__(identity=identity, types=types, **kwargs)
+            # Setting properties for GenBank's extraneous properties not settable in any SBOL3 field.
+            self.genbank_date          = sbol3.TextProperty(self, f"{self.GENBANK_EXTRA_PROPERTY_NS}#date"    , 0, 1)
```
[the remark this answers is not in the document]
mohitdmak: No, but these properties, at the minimum, are filled by GenBank with their defaults in case one does not provide explicit information on them while creating the GenBank SequenceRecords.

I've noticed there can be more "annotations" than these default ones, but couldn't find an exhaustive list of all possible annotations anywhere for this.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -131,20 +230,58 @@ def convert_genbank_to_sbol3(self, gb_file: str, sbol3_file: str = "sbol3.out",
...
                 types=COMP_TYPES + extra_comp_types,
                 roles=COMP_ROLES,
                 description=record.description,
             )
             doc.add(comp)
-            # NOTE: Currently we use a fixed method of encoding (IUPAC)
+            # Setting properties for GenBank's extraneous properties not settable in any SBOL3 field.
+            # 1. GenBank Record Date
+            # TODO: Let it be able to accept date into sbol3.DateTimeProperty() instead
+            comp.genbank_date = record.annotations['date']
```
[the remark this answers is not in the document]
mohitdmak: Yes, I am not sure how to store them. I created these properties explicitly as they are always provided by GenBank.
But as an eg, the test file in this PR has a "gi" annotation (see line 5 of the test GenBank file), which I did not anticipate based on the parsing of all GenBank files we have considered till now.

What should be our approach here for this? Should we add more properties as we encounter them in different GenBank files?

test/test_genbank_sbol3_direct.py
mohitdmak: Yep, thanks

mohitdmak: Right, I didn't notice this, fixed.

test/test_genbank_sbol3_direct.py
mohitdmak: Yea, it was not intended, fixed.

test/test_genbank_sbol3_direct.py
mohitdmak: Right, fixed this.

mohitdmak:
> Requested changes:
> 
> * The unit tests are failing
> * Remove extraneous top level files genbank.out, sbol.nt, and `test.gb.
> * There are a lot of upper case local variables and method arguments in test_genbank_sbol3_direct.py.

1. Fixed the unit test failures.
2. Apologies for the accidental check ins, removed them.
3. There were style issues with upper case variables in the test file, changed them apart along with those mentioned in the review.

> Comment, not to be fixed:
> 
> * convert_genbank_to_sbol3 and convert_sbol3_to_genbank are too long, and should be broken apart into a set of helper methods in a future pull request. Perhaps this should be a new GitHub issue for refactoring these methods.

I think there is one already created - https://github.com/SynBioDex/SBOL-utilities/issues/168

genbank.out
```diff
@@ -0,0 +1,133 @@
+LOCUS       AF165912                5485 bp    DNA     linear   PLN 29-JUL-1999
```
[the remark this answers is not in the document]
mohitdmak: Removed

tcmitchell:
> Also, @tcmitchell @jakebeal , if we are creating a seperate issue for the storing of References, should I remove all code related to storing references from this PR? Or should I only remove the previous attempts to store ReferenceLocations, and leave the rest property storage of Reference as is here?

You do not need to remove the code related to storing references from this PR. Leave whatever you think makes sense. It's also ok if some is stubbed out with an `if False`, for example, and a comment referring to the new issue as a TODO:

```python
    # TODO: See https://github.com/synbiodex/sbol-utilities/issue/4523
```

tcmitchell:
> I think there is one already created - https://github.com/SynBioDex/SBOL-utilities/issues/168

Yes! thanks, I missed that on a quick search, glad you found it.

> Apart from the above, there are 2 new issues to be created, as per your comments https://github.com/SynBioDex/SBOL-utilities/pull/158#issuecomment-1204081546 and https://github.com/SynBioDex/SBOL-utilities/pull/158#issuecomment-1204010036, and no commit pertaining to them is to be made in this PR right?

Correct. Please create those 2 new issues. You can link to specific comments in this issue, or cut/paste comments, or whatever makes sense for your workflow.

> There are a couple of points of discussion ongoing with @jakebeal in this PR, after which this would be ready to merge.

👍

> Post that, some commented blocks and debug statements would remain to be cleaned before the final merge.

👍

> Also, since we are creating a new issue for storing References, I have added a test file test/test_file/sbol3_genbank_conversion/test_extra_properties_with_references.gb which is not used currently but would be used to test the next PR related to the issue.

👍

mohitdmak:
> Correct. Please create those 2 new issues. You can link to specific comments in this issue, or cut/paste comments, or whatever makes sense for your workflow.

Alright, will update this comment, when I create them.

sbol_utilities/sbol3_genbank_conversion.py
jakebeal: I'd suggest moving both the logging and the main to a test, instead of making main run a test.

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
jakebeal: We need a way of handling an unexpected property.
I would suggest raising an exception for now.
Then we can open an exception for considering whether we want to do something more permissive.

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
jakebeal: As above, I suggest raising an exception for now, then opening an issue for considering whether to be more permissive.

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: Alright, removing them from the main file.

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
mohitdmak: Removed.

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
mohitdmak: Alright, I've created an issue here: https://github.com/SynBioDex/SBOL-utilities/issues/171

sbol_utilities/sbol3_genbank_conversion.py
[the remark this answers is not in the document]
mohitdmak: Done.

mohitdmak:
I have updated this PR and resolved the remaining reviews, and created issues on the topics we discussed in the last meeting. The PR desc is updated with recent commits and the issues.

tcmitchell (approved):
Looks good. Thanks for participating in the discussion around these changes, and for updating the code based on feedback.

PR #162, opened 2022-07-28
Synchronise with develop

mohitdmak:
A merge is needed to keep the feature branch for sbol3-genbank conversion updated with the develop branch. Recent fixes for unit-test issues due to rdflib upgrade requires the sync.

This PR makes no change to any component of the new converter.


tcmitchell (approved):
Thanks for setting this up.

PR #172, opened 2022-08-05
Supporting all GenBank files in test/test_files for round tripping

### LINKED ISSUE #165: Support all of the GenBank examples in test/test-files
Successfully round trip all of these files. (Note: there are only 2 files.)

This follows after #150, #164

mohitdmak:
This should be pretty easy to approve and merge
This should solve the issue: https://github.com/SynBioDex/SBOL-utilities/issues/165
**Note:** The large diff present currently will be reduced once the impending PR: https://github.com/SynBioDex/SBOL-utilities/pull/158 gets merged.

## Changes:
* This simply adds two new unit tests which check if the GenBank file stays identical in a round trip conversion from GenBank -> SBOL3 -> GenBank.
* The tests are added in `test/test_genbank_sbol3_direct.py`

Closes #165


test/test_genbank_sbol3_direct.py
```diff
@@ -133,6 +133,18 @@ def test_sbol3togb_2(self):
         )
         self._test_sbol3_to_genbank(sample_sbol3_file=sbol3_file, sample_genbank_file=genbank_file)
 
+    def test_round_trip_testfile_1(self):
```
tcmitchell, line 136: I'd like to see these two tests become a single test that iterates over all files with a `.gb` extension.

Typing from memory, so this may be a bit off, but should give you a feel for what I'm suggesting:

```python
test_file_dir = Path(__file__).parent / 'test_files'
for genbank_file in test_file_dir.glob('*.gb'):
    self._test_round_trip_genbank(genbank_file)
```

You will want to verify that it is in fact doing what I think it does, so you could print the file names temporarily to make sure.

This has the advantage that as genbank files are added to the `test_files` directory they will automatically be tested for successful round tripping.

tcmitchell (changes requested):

test/test_genbank_sbol3_direct.py
mohitdmak: Okay right, fixed this

test/test_genbank_sbol3_direct.py
```diff
@@ -150,6 +150,11 @@ def test_round_trip_multiple_loc_feat(self):
             Path(__file__).parent / "test_files" / "sbol3_genbank_conversion" / "multiple_feature_locations.gb"
         )
         self._test_round_trip_genbank(genbank_file)
+
+    def test_round_trip_all_testfiles(self):
+        test_file_dir = Path(__file__).parent / 'test_files'
+        for genbank_file in test_file_dir.glob('*.gb'):
+            self._test_round_trip_genbank(Path(genbank_file))
```
tcmitchell, line 157: Super sorry. This is quite minor. `pathlib.Path.glob` returns objects of type `pathlib.Path`, so you can just use `genbank_file` as is.

```suggestion
            self._test_round_trip_genbank(genbank_file)
```

tcmitchell (changes requested):
Terribly sorry. Just one very minor change is requested.

tcmitchell (approved):

PR #173, opened 2022-08-05
Ignore sbol objects that can't be parsed to GenBank objects

### LINKED ISSUE #164: Ignore SBOL objects that cannot be converted to Genbank
Skip over objects that cannot be converted to GenBank. I don't know if we want to log top level objects that are skipped, or if we should ignore them silently. 

This follows after #150

mohitdmak:
This solves #164 

## Changes:
* [Refactor] Removed some redundant test file paths defined in the main conversion file.
* Created a dictionary to keep logs of each top-level object and mark them as being parsed / not parsed during the conversion process.
* The dict returns the logs as well as the list of **SeqRecords**, nested in an outer dict.
* Modified the **convert_to_genbank** method of `sbol3_utilities/conversion.py` for recognizing the change in return value.

## TODO: 

Closes #164


mohitdmak:
@jakebeal @tcmitchell Here, I have implemented the basic idea discussed in #164, I am not sure if this needs refinement, and also how we are to write unit tests for it.

tcmitchell:
Could you write a test that passes a combination of objects in to the converter where at least one object is converted (result dict says `True`), and at least on object is not converted (result dict says `False`). Then check those results against the expectation that object A is converted and object B is not converted.

I don't know if that's possible. If not, we can think of other ways to test it. I think we should definitely have a test to verify that objects are properly either converted or not converted, and that the result dictionary accurately informs the caller about the status of each object.

mohitdmak:
> Could you write a test that passes a combination of objects in to the converter where at least one object is converted (result dict says `True`), and at least on object is not converted (result dict says `False`). Then check those results against the expectation that object A is converted and object B is not converted.


@tcmitchell  I have added a unit test to do the above stated. I added a test file with the standard component/sequences objects, along with an **Attachment** top-level object in the sbol3 file, which should return its conversion status as false.

tcmitchell (approved):

Issue #166, opened 2022-07-29
Support all of the sample files in the iGEM distribution


tcmitchell:
- Big, but not complicated
- 100s of GenBank conversions
- First step is to convert the GenBank files into SBOL and verify that the SBOL files are similar enough, although they don’t have to be identical
- Second step is to take the SBOL file that is produced, and convert that to a giant GenBank file and make sure that has substantially the same information as the current GenBank file that is produced

Note: This follows after #150, #164, #165

Issue #168, opened 2022-07-29
Refactor SBOL/Genbank converter to be more modular


tcmitchell:
The current SBOL3-Genbank converter is linear, one long function. More features and capabilities will be added. It is time to start thinking about how to break up this long function into parts, and have functions for each type of translation that is needed. This will allow the converter to grow more easily.

PR #176, opened 2022-08-24
add: ability to store and retrieve feature qualifiers

mohitdmak:
### Note:
* By the updates in this PR, complete, the converter can accurately round trip conversion for all iGEM distribution GenBank files of the form `BBF10K_000***.gb` (86 test files).
* A *leader* test file - `iGEM_BBF10K_000***.gb` has been added to `test/test_files/sbol3_genbank_conversion` to represent all related test files, and a unit test - `test_round_trip_iGEM_BBF10K` is added in place for it.
* The **only** case when absolute *memory-based* file diff would not return identical status is when either of the following is true:
    1. The **Keywords** field is not sorted in the given GenBank file. (see the 2nd TODO below)
    2. A few features are off of their correct position (based on *FeatureLocation*'s lexicographic sort) in the given GenBank file.

### Refactoring:
* Complete handling of **Sequence Features** creation and linking of corresponding **Feature Qualifiers** during conversion from GenBank to SBOL is now done by helper function `_handle_features_gb_to_sbol`.
* Similar handling of **Sequence Features** for conversion from SBOL to GenBank is done by helper function `_handle_features_sbol_to_gb`

### Updates:
* Added new test file `test/test_files/sbol3_genbank_conversion/feature_qualifier_storage.gb` to test storage and resetting of feature qualifiers, and updated a few other test files.
* New class `Feature_GenBank_Extension` extends the **sbol3.SequenceFeature** class, to include extra text properties of *Qualifier keys* and *Qualifier values*.
* A builder `build_feature_qualifiers_extension` is also registered for the same in the Converter's **__init__** to use this new class to parse **SequenceFeatures** encountered in an SBOL file.
* Many files in the iGEM distribution Suite have incorrect positioning of the *topology* information in the *data_file_division* section of the GenBank file. To resolve this, an extra check is added while parsing *topology* information:
```
topology = "linear"
if "topology" in record.annotations:
    topology = record.annotations["topology"]
# sometimes topology is specified in the 'data_file_division' field
elif record.annotations['data_file_division'] in ['circular', 'linear']:
    topology = record.annotations['data_file_division']
```
* Since the order of storage of *ListTextProperties* can get messed up while storing it in SBOL, in order to get the correct *Qualifier value* for a *Qualifier key*, we store the keys and not values not simply as their string formats, but with an id pre-pended by which we shall sort the array while resetting the qualifiers in conversion from SBOL to GenBank.
    - Eg: If Feature Qualifiers are `['qual_a': 'val_a', 'qual_b', 'val_b']`; they will be stored in extended sequence feature class's **Qualifier keys** as `['1: qual_a', '2: qual_b']` and in **Qualifier values** as `['1: val_a', '2: val_b']`. Thus, even if the order among keys and values is not preserved, we sort the items in both arrays by the integer before ':' and then iterate through them, ensuring that correct values are mapped to their respective keys.
```
keys = sorted(obj_feat.qualifier_key, key=lambda x: int(x.split(":", 1)[0]))
values = sorted(obj_feat.qualifier_value, key=lambda x: int(x.split(":", 1)[0]))
```

## TODO:
- [x]  Remove temporary iGEM distribution test files from PR.
- [x] Although, as per discussion on slack, sorting the list of **Keywords** before outputting Genbank files should be fine (see https://www.ncbi.nlm.nih.gov/Sitemap/samplerecord.html#KeywordsB), I am open to having the order maintained during conversion, using a similar workaround as we did for **Taxonomy** (storing the list as a string). This might reduce the amount of "false negatives" our converter test throws (A correct conversion is determined as incorrect due to different order of **Keywords**).


tcmitchell:
@jakebeal this PR adds a lot of iGEM genbank files. Do we want those added to this repository? Or should they be used by reference somehow, like downloading them from GitHub during testing, or making the iGEM repo a submodule, or some other approach. I assume the iGEM genbank files will change over time.

mohitdmak:
@tcmitchell  @jakebeal, I have implemented the storage and resetting of feature qualifiers as per our discussion in the last meeting; this should be good to merge unless there are any issues with the review.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -213,53 +253,9 @@ def convert_genbank_to_sbol3(self, gb_file: str, sbol3_file: str = "sbol3.nt", n
             # Setting properties for GenBank's extraneous properties not settable in any SBOL3 field.
             self._store_extra_properties_in_sbol3(doc, comp, seq, record)
 
-            if record.features:
```
tcmitchell, line 216: The extraction of this code block to a separate method is really great, and a nice step towards #168

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -339,83 +334,9 @@ def convert_sbol3_to_genbank(self, sbol3_file: str, doc: sbol3.Document = None,
                 self._reset_extra_properties_in_genbank(obj, seq_rec, references)
 
                 seq_rec_features = []
-                if obj.features:
```
tcmitchell, line 342: The extraction of this code block to a separate method is really great, and a nice step towards #168

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -586,3 +510,157 @@ def _reset_extra_properties_in_genbank(self, obj: sbol3.Component, seq_rec: SeqR
...
+    
+    def _handle_features_gb_to_sbol(self, record: SeqRecord, comp: sbol3.Component, seq: sbol3.Sequence) -> None:
+        """Helper function for setting sequence features and their qualifiers to SBOL,
+        by using a modified, extended SBOL3 Sequence Feature class - Feature_GenBank_Extension.
+        :param record: GenBank SeqRecord instance for the record which contains sequnce features
+        :param comp: Instance of the SBOL3 Component 
+        :param seq: The Sequence used in the GenBank record corresponding to sbol3 comp
+        """
+        # parse if genbank record has any features
+        if record.features:
```
tcmitchell, line 522: Note: I am not asking for a change, I am making an observation.

After reviewing the code for the case where there are `record.features` I scrolled down to see what happens when there are no record features. I think the answer is that nothing happens, but it was hard to tell. A different option is to short-circuit the method and de-indent the body:

```python
if not record.features:
    # There is nothing to do if there are no features
    return
comp.features = []
for ind, gb_feat in ....
    ...
# rest of method goes here
```

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -586,3 +510,157 @@ def _reset_extra_properties_in_genbank(self, obj: sbol3.Component, seq_rec: SeqR
...
+
+    def _handle_features_sbol_to_gb(self, seq_rec: SeqRecord, obj: sbol3.Component, seq_rec_features: List) -> None:
+        """Helper function for resetting sequence features and their qualifiers to GenBank,
+        by using a modified, extended SBOL3 Sequence Feature class - Feature_GenBank_Extension.
+        :param seq_rec: GenBank SeqRecord instance for the record which contains sequnce features
+        :param obj: Instance of the SBOL3 Component 
+        :param seq_rec_features: List of features for the related sequence features
+        """
+        seq_rec_features = []
+        if obj.features:
```
tcmitchell, line 588: Note: I am not asking for a change, I am making an observation.

The same applies here as in `_handle_features_gb_to_sbol`. This method could short-circuit the case where there are no `obj.features`. It is hard for me to tell if the last two lines of the method align with this `if` statement, or if they are indented inside of the `if` statement.

tcmitchell (approved):
Overall, nice work. I made a few comments, and no changes are needed or requested.

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: @tcmitchell Thanks for pointing this out! I have inculcated this.

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: @tcmitchell  same here, have made this change as it does improve the readability, thanks for pointing it out

mohitdmak:
@tcmitchell  @jakebeal  I have included the storage of Keywords as a string instead of a sorted list, and also a couple of small refactors as suggested by Tom, this PR should be good to merge.

tcmitchell:
* I think this PR is related to issue #166 (Support all of the sample files in the iGEM distribution). Is that correct?
* Does this PR close #166 or is it another step on the path to #166?

mohitdmak:
> * I think this PR is related to issue [Support all of the sample files in the iGEM distribution #166](https://github.com/SynBioDex/SBOL-utilities/issues/166) (Support all of the sample files in the iGEM distribution). Is that correct?

Yes
> * Does this PR close [Support all of the sample files in the iGEM distribution #166](https://github.com/SynBioDex/SBOL-utilities/issues/166) or is it another step on the path to [Support all of the sample files in the iGEM distribution #166](https://github.com/SynBioDex/SBOL-utilities/issues/166)?

Yes, it adds support for a big chunk of those files, only a few remain now

PR #181, opened 2022-09-21
add: support for remaining iGEM testsuite files

mohitdmak:
### Comment / Structured Comment Storage
* Certain iGEM test suite files had record annotations of "**comment**" (*text*) and "**structured_comment**" (*dictionary of dictionaries*).
* The class `CustomStructuredCommentProperty` is a *CustomTopLevel* extended class, to support setting and resetting of **structured_comment** in a round trip conversion of GenBank. It has a corresponding builder `build_custom_structured_comment_property` registered and invoked for parsing it while reading any SBOL3 file in the main `GenBank_SBOL3_Converter` class's `__init__()` call.
* The process of setting and resetting the **structured_comment** property is followed in a similar way to that followed for **FeatureQualifiers**. The inner dictionary's keys and values are stored in **TextListProperty** along with the index in which they occur in the original GenBank file - these key and value arrays are then sorted by this index before resetting them in GenBank files.
* Before the resetting of these **structured_comments**, a dictionary is created to link each SBOL3 component with the structured comments written under the corresponding sequence record in the original GenBank file. This has to be done as the `CustomStructuredCommentProperty` is a top-level class, and is not directly linked with each SBOL3 component.
* The **comment** text is stored in an added *TextProperty* of the extended component class `Component_GenBank_Extension`. It's setting and resetting is handled by the helper functions `_reset_extra_properties_in_genbank` and `_store_extra_properties_in_sbol3`.

### DB Links Storage
* Certain files had a list of texts for their "**dbxrefs**" properties on Sequence Records. Storage of these is implemented in a similar manner to **taxonomy**, where the list is stored as a string in the *genbank_dblink* property of the `Component_GenBank_Extension` class. Upon conversion to GenBank, the string is split into a list. This is again handled by the helper functions `_reset_extra_properties_in_genbank` and `_store_extra_properties_in_sbol3`.

### New test / test files
* New test file `test_structured_comments.gb` has been added to unit test the setting and resetting of structured comments in a round trip. This file is exactly the same as the `NCBI_GenBank_imports.gb` test file in the iGEM Test suite.

## TODO:

- [x]  A sample unit test needs to be added for files having comment / structured properties.
- [x]  A sample unit test needs to be added for files having DB Link properties.

Related to #166

### DIFF
--- a/sbol_utilities/sbol3_genbank_conversion.py
+++ b/sbol_utilities/sbol3_genbank_conversion.py
@@ -34,9 +34,6 @@ class GenBank_SBOL3_Converter:
     DEFAULT_GB_TERM = "misc_feature"
     # Namespace to be used be default if not provided, and also for all unit tests related to this converter
     TEST_NAMESPACE = "https://test.sbol3.genbank/"
-    CUSTOM_REFERENCE_PROPERTY_URI = "http://www.ncbi.nlm.nih.gov/genbank#reference"
-    FEATURE_QUALIFIER_PROPERTY_URI = "http://www.ncbi.nlm.nih.gov/genbank#featureQualifier"
-    CUSTOM_STRUCTURED_COMMENT_PROPERTY_URI = "http://www.ncbi.nlm.nih.gov/genbank#structured_comment"
     # File locations for required CSV data files which store the ontology term translations between GenBank and SO ontologies
     GB2SO_MAPPINGS_CSV = os.path.join(os.path.dirname(os.path.realpath(__file__)), "gb2so.csv")
     SO2GB_MAPPINGS_CSV = os.path.join(os.path.dirname(os.path.realpath(__file__)), "so2gb.csv")
@@ -96,9 +93,9 @@ class GenBank_SBOL3_Converter:
         # the SBOL3 parser to build objects with a Component type URI
         sbol3.Document.register_builder(sbol3.SBOL_COMPONENT, build_component_genbank_extension)
         # Register the buildre function for custom reference properties
-        sbol3.Document.register_builder(self.CUSTOM_REFERENCE_PROPERTY_URI, build_custom_reference_property)
+        sbol3.Document.register_builder(self.Feature_GenBank_Extension.GENBANK_FEATURE_QUALIFIER_NS, build_custom_reference_property)
         # Register the buildre function for custom structured comment properties
-        sbol3.Document.register_builder(self.CUSTOM_STRUCTURED_COMMENT_PROPERTY_URI, build_custom_structured_comment_property)
+        sbol3.Document.register_builder(self.CustomStructuredCommentProperty.CUSTOM_STRUCTURED_COMMENT_NS, build_custom_structured_comment_property)
         # # Register the builder function so it can be invoked by
         # # the SBOL3 parser to build objects with a SequenceFeature type URI
         sbol3.Document.register_builder(sbol3.SBOL_SEQUENCE_FEATURE, build_feature_qualifiers_extension)
@@ -318,7 +315,7 @@ class GenBank_SBOL3_Converter:
         # create dict to link component with their respective Reference property objects
         references: Dict[sbol3.Component, List[sbol3.CustomTopLevel]] = {}
         for obj in doc.objects:  
-            if isinstance(obj, sbol3.CustomTopLevel) and obj.type_uri == self.CUSTOM_REFERENCE_PROPERTY_URI:
+            if isinstance(obj, sbol3.CustomTopLevel) and obj.type_uri == self.CustomReferenceProperty.CUSTOM_REFERENCE_NS:
                 component_object = doc.find(str(obj.component))
                 if component_object and isinstance(component_object, sbol3.Component):
                     references[component_object] = [obj] if component_object not in references else references[component_object] + [obj]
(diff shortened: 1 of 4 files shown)


sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -113,6 +126,22 @@ def __init__(self, type_uri=CUSTOM_REFERENCE_NS, identity="customReferenceProper
             self.location = sbol3.OwnedObject(self, f"{self.CUSTOM_REFERENCE_NS}#location", 0, math.inf, type_constraint=sbol3.Range)
 
 
+    class CustomStructuredCommentProperty(sbol3.CustomTopLevel):
+        """Serves to store information and annotations for 'Structured_Comment' objects in 
+        GenBank file to SBOL3 while parsing so that it may be retrieved back in a round trip
+        :extends: sbol3.CustomTopLevel class
+        """
+        CUSTOM_STRUCTURED_COMMENT_NS = "http://www.ncbi.nlm.nih.gov/genbank#structured_comment"
+        def __init__(self, type_uri=CUSTOM_STRUCTURED_COMMENT_NS, identity="customStructuredCommentProperty"):
```
tcmitchell, line 135: I think this should use the defined constant `CUSTOM_STRUCTURED_COMMENT_PROPERTY_URI` instead of defining a new constant. But maybe the two represent different things and it is coincidental that their values are the same. Use your judgment here, I may be incorrect about this.

tcmitchell (approved):
Everything looks good to me. I just had a minor question about two constants that appear to have the same value.

tcmitchell:
@mohitdmak will:

* add the sample unit test for files having DB Link properties
* will refactor to handle the duplication of the value held by `CUSTOM_STRUCTURED_COMMENT_PROPERTY_URI` mentioned in the review comment above

Then we will re-review.

mohitdmak:
@tcmitchell @jakebeal  recent commits have added the required unit test and removed redundant namespace constants from the outer class. This PR is good to merge now.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -428,6 +468,26 @@ def _store_extra_properties_in_sbol3(self, doc: sbol3.Document, comp: Component_
                     # TODO: Raise error, no name for component
                     # else:
                     doc.add(custom_reference)
+            # 14. GenBank Record Comment
+            elif annotation == 'comment':
+                comp.genbank_comment = record.annotations['comment']
+            # 15. GenBank Record Structured comments
```
tcmitchell, line 474: These comments are very helpful!

test/test_genbank_sbol3_direct.py
```diff
@@ -209,6 +209,24 @@ def test_round_trip_iGEM_BBF10K(self):
         )
         self._test_round_trip_genbank(genbank_file)
     
+    def test_round_trip_structured_comments(self):
+        """Test ability to correctly round trip genbank test files in the iGEM distribution which have 
+        comment and structured comment annotations.
+        """
+        genbank_file = (
+            Path(__file__).parent / "test_files" / "sbol3_genbank_conversion" / "test_structured_comments.gb"
```
tcmitchell, line 217: In a future world, perhaps by someone else, we could create a variable to hold the `Path` to the `sbol3_genbank_conversion` directory. In general, watch for these patterns and extract them to common code so they are easier to update in the future.

```python
TEST_DIR = Path(__file__).parent / "test_files" / "sbol3_genbank_conversion"
```

Note: _This is not a requested change, just a remark on the code._

tcmitchell (approved):

tcmitchell:
@jakebeal the code looks fine to me. Please review when you are able from the perspective of iGEM. Thanks!

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -359,6 +397,9 @@ def _store_extra_properties_in_sbol3(self, doc: sbol3.Document, comp: Component_
         :param record: GenBank SeqRecord instance for the record which contains extra properties
         """
         comp.genbank_record_id = record.id
+        # set dblinks from the dbxrefs property of biopython
+        if record.dbxrefs:
+            comp.genbank_dblink = "::".join(record.dbxrefs)
```
jakebeal, line 402: I don't understand what this is, and the comment above isn't helping me. Can you please expand the comment to explain the purpose, maybe link to a test example as well?
I'm pretty sure this is about structured comments, but there's nothing in the code here that's explaining that.

sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -113,6 +123,22 @@ def __init__(self, type_uri=CUSTOM_REFERENCE_NS, identity="customReferenceProper
             self.location = sbol3.OwnedObject(self, f"{self.CUSTOM_REFERENCE_NS}#location", 0, math.inf, type_constraint=sbol3.Range)
 
 
+    class CustomStructuredCommentProperty(sbol3.CustomTopLevel):
+        """Serves to store information and annotations for 'Structured_Comment' objects in
```
jakebeal, line 127: I think it would be good for the docstring to include a pointer to NCBI's reference page on structured comments: https://www.ncbi.nlm.nih.gov/genbank/structuredcomment/

jakebeal (approved):

jakebeal:
I am approving, with two minor requests for improved documentation that I think can be handled before the merge.

test/test_genbank_sbol3_direct.py
mohitdmak: Right thanks, this makes sense

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: I have put up the link in the docstring in the latest commit

sbol_utilities/sbol3_genbank_conversion.py
mohitdmak: My bad, I forgot to explain this; I have added comments in the latest commit; this is just about how biopython stores dbxrefs in a list, and how '::' is used as a delimiter to store the list as a flat string and then reset back to a list during round trip conversion.

mohitdmak:
@tcmitchell @jakebeal  I have resolved the pr reviews, and documentation, I suggest merging this PR soon so that the diff on https://github.com/SynBioDex/SBOL-utilities/pull/183 becomes readable. (It includes the changes also of this pr)

tcmitchell (approved):

PR #186, opened 2022-10-19
final merge: Offline SBOL3 <-> GenBank Converter; Integration with previous online Converter

mohitdmak:
## TODO:


jakebeal:
I have begun testing with the iGEM distribution, and run into an error. It appears that converted GenBank files are getting a 0-based "RangeStart" value rather than 1-based, which is both incorrect and results in invalid SBOL when the range should start at 1. An example of such a file:

[package.nt.txt](https://github.com/SynBioDex/SBOL-utilities/files/9845178/package.nt.txt)

This was made by converting GenBank files such as the following:
[111772.gb.txt](https://github.com/SynBioDex/SBOL-utilities/files/9845179/111772.gb.txt)

jakebeal (changes requested):
The 0-based range error needs to be corrected.

mohitdmak:
> The 0-based range error needs to be corrected.

This turned out to be an issue as BioPython parsed the start Locations of GenBank files based on 0-indexing instead of 1-indexing. I have created a [PR](https://github.com/SynBioDex/SBOL-utilities/pull/188) for the fix, where we convert the start value to 1-indexed when converting from GenBank to SBOL3 and 0-indexed for the opposite direction.

jakebeal:
It appears that something changed between biopython 1.79 and 1.81 that is causing the tests to fail. This will need investigation.

tcmitchell:
Hey @jakebeal can this one be merged now? The tests are all passing after fixing the BioPython Hashable change.

jakebeal:
I am working through some errors that come up in its usage with the iGEM workflow. Once all of those are either resolved or sufficiently identified to defer into other tickets, I will merge this.

jakebeal:
Once PR #203 and PR #206 are merged, this branch will be ready to merge.

There is still one major outstanding issue that will need to be addressed, which is the "flattening" of hierarchical Component objects to include the features of their SubComponents. I have filed that as #207, and it can be addressed separately from this PR

Issue #185, opened 2022-10-17
find_feature


jakebeal:
In the SBOL tutorial, we use a loop for searching for a feature:
```
for f in i13504.features:
    if f.instance_of == b0015.identity:
        print(f)
        f.locations.append(Range(i13504_seq, 746, 875))
```
The search through features could be turned into a little macro routine, e.g., `def find_feature(matching: Callable) -> sbol3.Feature`

PR #195, opened 2023-02-04
find feature function 

ahmedtarek26:
Instead of using loops, this function is to search for a specific feature in the sbol3 file.
I added the function in the `helper_function.py`, I think this is a good place to put this function, also we can add the function in the community media repository.


sbol_utilities/helper_functions.py
```diff
@@ -357,3 +357,9 @@ def collector(obj: sbol3.Identified):
     with cached_references(doc):
         doc.traverse(collector)
     return outgoing
+
+def find_feature(sbol3: sbol3_file,f: feature,seq: sequence) -> sbol3.Feature:
```
jakebeal, line 361: Can you please add a docstring that documents the behavior of this function, following the same pattern as the other functions in the file?

It also appears that your type hints are incorrect; does this run correctly when you test it?

sbol_utilities/helper_functions.py
```diff
@@ -357,3 +357,9 @@ def collector(obj: sbol3.Identified):
...
+    for i in sbol3.features:
+        if i.instance_of == f.identity:
+            i.locations.append(Range(seq, 746, 875))
```
jakebeal, line 364: Why are you appending the Range here?

jakebeal:
This function needs to have a test added that validates its behavior before it can be merged.

jakebeal (changes requested):

sbol_utilities/helper_functions.py
```diff
@@ -357,3 +357,12 @@ def collector(obj: sbol3.Identified):
     with cached_references(doc):
         doc.traverse(collector)
     return outgoing
+
+def find_feature(doc: sbol3.Component,f:sbol3.Feature) -> sbol3.Feature:
+    """
+    find_feature function used for  finding a specific feature you specify and check
+    if this feature one of your sbol3 document features or not.
```
jakebeal, line 364: It is not a document, but a component. You should change the variable name as well.

test/test_helpers.py
```diff
@@ -136,6 +136,29 @@ def test_outgoing(self):
...
+        doc.add(i13504)
+        # creating a feature and add it to the sbol3 Document
+        b0034 = sbol3.Component('B0034', sbol3.SBO_DNA)
+        b0034.name = 'RBS (Elowitz 1999)'
+        b0034.roles = [sbol3.SO_NS + '0000139']
+        doc.add(b0034)
+        i13504.features.append(sbol3.SubComponent(b0034))
+
+        result=find_feature(i13504,b0034)
+        self.assertEqual(find_feature(i13504,b0034),result)
```
jakebeal, line 158: This test does not actually check if the function is operating correctly. How would you know if result is actually the right result?

sbol_utilities/helper_functions.py
```diff
@@ -357,3 +357,12 @@ def collector(obj: sbol3.Identified):
     with cached_references(doc):
         doc.traverse(collector)
     return outgoing
+
+def find_feature(doc: sbol3.Component,f:sbol3.Feature) -> sbol3.Feature:
```
jakebeal, line 361: This isn't actually addressing #185, as one has to already have the feature in order to find it.
In the issue, the example is looking for a feature whose identity is known, but not that actual object.

sbol_utilities/helper_functions.py
```diff
@@ -357,3 +357,12 @@ def collector(obj: sbol3.Identified):
...
+    """
+    find_feature function used for  finding a specific feature you specify and check
+    if this feature one of your sbol3 document features or not.
+    """
+    for i in doc.features:
+        if i.instance_of == f.identity:
```
jakebeal, line 367: This will never return true.

sbol_utilities/helper_functions.py
```diff
@@ -357,3 +357,12 @@ def collector(obj: sbol3.Identified):
...
+            return i
```
jakebeal, line 368: What should happen if you don't find the feature?

jakebeal (changes requested):

sbol_utilities/helper_functions.py
ahmedtarek26: yeah, got it

what if I used this line of code? 
` if i.instance_of.split('/')[-1] == f.identity.split('/')[-1]:`
using `.identity` with the feature we input and find the feature name then comparing it with the same feature if it exist using `.instance_of`

jakebeal: That doesn't address the question of the function's interface.

tcmitchell:
@jakebeal Please note that this PR is targeted to `main` instead of `develop`. That might not be what you want.

jakebeal:
As no updates have been made within the last several months to address the requested changes, I am going to close this PR.

Issue #198, opened 2023-03-08
TypeError: unhashable type: 'SeqFeature'


tcmitchell:
The sbol3-genbank conversion is failing with "TypeError: unhashable type: 'SeqFeature'" errors. This appears to be a change in BioPython 1.80, released 18 November 2022.

Here is a simple test with BioPython 1.79:
```python
>>> import Bio
>>> Bio.__version__
'1.79'
>>> from Bio.SeqFeature import SeqFeature
>>> x = SeqFeature()
>>> tbl = {}
>>> tbl[x] = 'b'
>>> 
```

Here is the same test with BioPython 1.80:
```python
>>> import Bio
>>> Bio.__version__
'1.80'
>>> from Bio.SeqFeature import SeqFeature
>>> x = SeqFeature()
>>> tbl = {}
>>> tbl[x] = 'b'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'SeqFeature'
>>> 
```

tcmitchell:
There are 12 unit test errors and all 12 are caused by the same issue. That's good, one fix is probably all that is needed.

Function `_handle_features_sbol_to_gb` builds a `dict` whose key is a `SeqFeature` and whose value is a list of lists of start/end pairs. I suspect we simply have to come up with a different representation to store this association. Using a `SeqFeature` as a hashable key in a `dict` worked in BioPython 1.79, and stopped working in 1.80.

PR #199, opened 2023-03-08
Fix SeqFeature not hashable error

tcmitchell:
SeqFeature is no longer hashable as of BioPython 1.80. Try an alternate implementation that avoids using SeqFeature instances as keys in a dict. Store the loc_positions on a new property of SeqFeature. Akin to monkey patching.

### DIFF
--- a/sbol_utilities/sbol3_genbank_conversion.py
+++ b/sbol_utilities/sbol3_genbank_conversion.py
@@ -894,7 +894,7 @@ class GenBank_SBOL3_Converter:
                 feat = SeqFeature(
                     location=feat_loc_object, strand=feat_strand, type=feat_role
                 )
-                feat_order[feat] = feat_loc_positions
+                feat.loc_positions = feat_loc_positions
                 if isinstance(obj_feat, self.Feature_GenBank_Extension):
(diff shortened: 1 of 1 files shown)


sbol_utilities/sbol3_genbank_conversion.py
```diff
@@ -904,6 +904,6 @@ def _handle_features_sbol_to_gb(self, seq_rec: SeqRecord,
 
         # Sort features based on feature location start/end, lexicographically, and then by
         # strand / number of qualifiers / type of feature string comparison
-        seq_rec_features.sort(key=lambda feat: (feat_order[feat], feat.strand, len(feat.qualifiers), feat.type))
```
jakebeal, line 907: Is the feat_order variable now used anywhere? If not, then it should probably be removed too.

jakebeal (approved):

jakebeal:
Nice debugging! Approved with one style improvement suggestion to resolve.

sbol_utilities/sbol3_genbank_conversion.py
tcmitchell: Good catch. I have removed the variable in a new commit.

jakebeal:
Note: this fixed #198

PR #202, opened 2023-03-27
iGEM Quality Checker System

rkrishnasanka:
(Will update description when completed)

Notes:
- Currently the qc checker does not parse the data from the excel sheet / anywhere. 
- The data needs to passed to the checker in the form of a dictionary (can be converted into a datastructure, with dictionary-esque data stored for each of the parameters). This we cannot avoid if we don't want the QC checker to pull the data on its own.
- Currently it doesn't check if the data is the right sort for SBOL, maybe we need to talk about SBOL validation (either calling a method or specifying a validation derived from the SBOL library)
- Still need to create the unit tests and integration test to test the whole validation pipeline.

Updated:

I realized the easiest way to validate the parsed data would be to manually test against an entity. Here's the API:

```
import json
from pathlib import Path
from sbol_utilities.igem.qc_checker import QCChecker


data_dict = {
    "field_1": {},
    "field_2": {},
}


json_dict = None
with open('qc_schema.json') as json_file:
    json_dict = json.load(json_file)
checker = QCChecker.from_json(json_dict)
checker.perform_qc_check("top_level_entity_1", data_dict)

```

By having the entity id being passed to the QC checker, we simplify the validation logic to be couple with the parsing logic. 

I can add an additional parser for Excel row/column selection


rkrishnasanka:
@jakebeal do you want to take a look at these notes ^^^

PR #144, opened 2022-07-16
Create calculate_complexity_score.py

jguillen15:
First attempt to create SBOL utility to calculate complexity score of a sequence

PR #211, opened 2023-05-16
Complexity score utility

jakebeal:
This PR contains new utility, created by @jguillen15, that uses IDT's API to calculate synthesis complexity scores for sequences. High scores are bad, low scores are readily able to be synthesized.

The initial draft was brought into the repository via Pull Request #144, and has since been cleaned up to production-ready by me.


setup.py
```diff
@@ -41,6 +41,8 @@
             'graphviz',
             'tyto>=1.4',
             'openpyxl',
+            'certifi',
```
tcmitchell: I don't think `certify` needs to be included here. `requests` depends on it, but this utility does not depend directly on `certifi`. You can let `requests` bring in `certifi`, sbol-utilities doesn't have to.

tcmitchell (approved):

setup.py
jakebeal: Thank you; you were right that deleting it did not hurt anything.

PR #213, opened 2023-07-25
Fix small issue on line 230 in calculate_complexity_scores.py

jguillen15:
There is an issue with the if statement on line 230. This statement tests whether "credentials" is a key in the dictionary (i.e., is it an argument), which is always true.  It should instead be checking the value that is assigned to that key.


jakebeal:
Tests are failing - please identify the problem and fix it so that the tests pass.

jakebeal (approved):

jakebeal:
The tests are succeeding in the source fork. Their failure here appears to be due to GitHub's protection against sharing secrets across forks. Accordingly I'm going to try merging and see if that fixes it...

PR #214, opened 2023-10-06
Change filename with star to stop Windows failures

jakebeal:
Filenames with star in them fail on some Windows machines, so rename the test file with stars in its name to not have stars


jakebeal:
Checks passed, and I'm going to merge without review for this emergency patch.

PR #215, opened 2023-10-07
First steps of SBOL3 to SBOL2 conversion in Python

jakebeal:
This pull request provides a framework for implementing SBOL3/SBOL2 conversion in Python, rather than via a call-out to JavaScript.

The current converters from SBOL2 to SBOL3 and from SBOL3 to SBOL2 are given an option that currently defaults to false, but when set to true will convert using the Python package instead of calling JavaScript.

I have provided a full stub-out of the SBOL3 to SBOL2 conversion, and have implemented conversion of the following classes:

* Identified
* TopLevel
* Sequence
* Activity (except for object references)
* Component (except for object references)


PrashantVaidyanathan (approved):
This is a very useful starting point. Looks good to me.

PR #217, opened 2023-10-07
Round-trip of SBOL2/SBOL3 python conversion

jakebeal:
Clean up minor details to allow triple-to-triple identity in first tests of SBOL2/SBOL3 Python conversion

Note that pySBOL2 validation fails, apparently due to an issue with SHACL rules for the DateTime type (https://github.com/SynBioDex/pySBOL2/issues/429)


jakebeal:
I am going ahead with a merge as there has been no review for several days.

PR #224, opened 2023-11-08
Implemented converstion of collection between SBOL2 and SBOL3 for the…

Yehuda-Binik:
In the new converter, I completed the `visit_collection` function for both the `SBOL3To2ConversionVisitor` and `SBOL2To3ConversionVisitor` classes. I also added two test cases to test my code using the SBOL3 [collection.nt](https://github.com/SynBioDex/SBOLTestSuite/blob/master/SBOL3/entity/collection/collection.nt) file from the [SBOLTestSuite](https://github.com/SynBioDex/SBOLTestSuite/tree/master).


jakebeal (approved):
Approving despite test failure, because test failure is unrelated, caused by missing secrets for another test.

Issue #205, opened 2023-04-08
Ensure all GenBank features are in term translation list


jakebeal:
The official list of supported GenBank features may be found at: https://www.insdc.org/submitting-standards/feature-table/#7.2

Most, but not all, are in the `gb2so.csv` and `so2gb.csv` translation maps with equivalent Sequence Ontology terms; make sure that all are there.

arnav-luhadiya:
Hi @jakebeal. What if the equivalent Sequence Ontology term of the feature is not defined. 
Should I not write the feature itself or just leave the `SO` equivalent term?

jakebeal:
We want to make sure that every GenBank feature has a corresponding SO term.

Given that SO has a much richer vocabulary, I suspect that every feature should map to a term. If there are features that don't appear to have a corresponding term, or that are ambiguous, however, it should be discussed on this issue. If necessary, we can also put in placeholders and ask for new SO terms.

arnav-luhadiya:
I could not find corresponding term of `old_sequence` for example. 
Checked the following links :
[Sequence Ontology Search](http://www.sequenceontology.org/browser)
[Sequence Ontology Mapping](http://www.sequenceontology.org/resources/mapping/FT_SO.html)

Attaching a relevant screenshot from the (Sequence Ontology Mapping) link:

![image](https://github.com/SynBioDex/SBOL-utilities/assets/114944869/aa226772-d170-4a71-8c47-e9f6c84dd84b)

arnav-luhadiya:
> I could not find corresponding term of `old_sequence` for example. Checked the following links : [Sequence Ontology Search](http://www.sequenceontology.org/browser) [Sequence Ontology Mapping](http://www.sequenceontology.org/resources/mapping/FT_SO.html)
> 
> Attaching a relevant screenshot from the (Sequence Ontology Mapping) link:
> 
[...]

PR #226, opened 2024-01-16
Genbank features in translation list

arnav-luhadiya:
This is in response to issue #205.
Added the GenBank features which were not in the translation list. Also updated corresponding `SO` terms for some features according to [Sequence Ontology Mapping](http://www.sequenceontology.org/resources/mapping/FT_SO.html).

Also, there seem to be instances of the same feature with multiple `SO` terms in `so2gb.csv`. Do you want to keep all of them or only one unique `SO` term for each feature as given in `gb2so.csv` file?

### DIFF
--- a/sbol_utilities/conversion.py
+++ b/sbol_utilities/conversion.py
@@ -156,7 +156,7 @@ def convert2to3(sbol2_doc: Union[str, sbol2.Document], namespaces=None, use_nati
     # remap orientation types
     orientation_remapping = {
         sbol2.SBOL_ORIENTATION_INLINE: sbol3.SBOL_INLINE,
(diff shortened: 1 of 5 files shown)


jakebeal:
Can you please rebase the commits for this request on develop, rather than on your other branch? As it currently is, we cannot merge this branch without also merging the other branch.

This has also added another test failure:
```
test/test_genbank_sbol3_direct.py::TestGenBankSBOL3::test_feature_location_types_ignore_fuzzy
```

jakebeal:
As it has been more than two months without an update to address the issues in the pull request, I am going to decline it for now. @arnav-luhadiya , if you are able to address the issues, please feel free to set up a new pull request with an updated version of your modifications.

PR #236, opened 2024-04-01
Enhanced Error Handling for Missing Columns in Excel Input

### LINKED ISSUE #46: Opaque error when data sources is missing a column
When the "data sources" tab is missing a column (e.g., column G on old versions of the sheet), it exceptions with an opaque error:
```
>>> doc = es.excel_to_sbol(wb)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python39\site-packages\sbol_utilities\excel_to_sbol.py", line 426, in excel_to_sbol
    basic_parts, composite_parts, linear_products, final_products, source_table = read_metadata(wb, doc, config)
  File "C:\Python39\site-packages\sbol_utilities\excel_to_sbol.py", line 110, in read_metadata
    source_table = {row[config['source_name_col']].value: row[config['source_uri_col']].value
  File "C:\Python39\site-packages\sbol_utilities\excel_to_sbol.py", line 112, in <dictcomp>
    if row[config['source_literal_col']].value}
IndexError: tuple index out of range
```
It should at least be more graceful about its crash.

vinayakjeet:
#46 

## Problem Description
When the "data sources" tab in the provided Excel files is missing a column, such as column G in older versions of the spreadsheet, the script encounters an unhandled exception and crashes. This results in an opaque error message that does not clearly indicate the cause of the problem, leading to potential confusion and making troubleshooting difficult.

## Proposed Solution
To address this issue, it is suggested to wrap the section of the code that accesses potentially missing columns in a `try-except` block. This block should catch `IndexError` exceptions and log a warning that clearly identifies the missing column and provides guidance or suggestions for resolving the issue. The script should then continue to execute, allowing for the processing of any additional data that does not depend on the missing columns.

## Expected Benefits
- **Improved Resilience:** By catching and handling `IndexError` exceptions, the script will no longer crash when encountering missing columns, improving its overall resilience.
- **Clearer Error Reporting:** The logging of specific warnings when columns are missing will make troubleshooting more straightforward, enhancing the user experience.
- **Enhanced Script Usability:** Allowing the script to continue execution after encountering a missing column ensures that users can still obtain partial results from their input files, improving the script's usability in less-than-ideal scenarios.

## Additional Notes
Implementing this solution does not prevent the issue of missing columns but provides a more graceful handling mechanism that benefits users by improving error transparency and script robustness.

Closes #46

### DIFF
--- a/sbol_utilities/excel_to_sbol.py
+++ b/sbol_utilities/excel_to_sbol.py
@@ -107,9 +107,13 @@ def read_metadata(wb: openpyxl.Workbook, doc: sbol3.Document, config: dict):
     doc.add(final_products)
 
     # also collect any necessary data tables from extra sheets
-    source_table = {row[config['source_name_col']].value: row[config['source_uri_col']].value
-                    for row in wb[config['sources_sheet']].iter_rows(min_row=config['sources_first_row'])
-                    if row[config['source_literal_col']].value}
+    source_table = {}
+    try:
+        source_table = {row[config['source_name_col']].value: row[config['source_uri_col']].value
+                        for row in wb[config['sources_sheet']].iter_rows(min_row=config['sources_first_row'])
+                        if row[config['source_literal_col']].value}
+    except IndexError as e:
+        logging.warning(f"An error occurred while reading the sources sheet, possibly due to a missing column: {e}")
 
     # return the set of created collections
     return basic_parts, composite_parts, linear_products, final_products, source_table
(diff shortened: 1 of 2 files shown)


jakebeal (changes requested):
This pull request mixed in the described change with a number of other changes in the `helper_functions.py` file. There are a number of failing tests, likely caused by these other changes.

There is also not currently a new test added to demonstrate that the exception is generated and caught as intended.

Can you please address these issues in this pull request?

PR #260, opened 2024-04-22
Issue 249

manulera:
cc: @jakebeal 

I tried, but I don't think I made any meaningful progress. Leaving the code here in case it would help the next person.

I think the visit_interaction and visit_participant might be useful for the next person (see python scripts added, specifically compare_participation.py), but I struggled with higher order classes. The way the tests are structured requires adding everything to a document, so all the converters for higher-order classes have to be written in order to test the lower order classes. I see why this lack of modularity happens (Ids are assigned when adding to a top-level), but this makes it very hard to start with the easiest classes (not top-level).

Reading the spec I could not figure out how to map the multiple SBOL2 classes that map to SBOL3 component and subcomponent. Not having used SBOL in the past, I don't think I can contribute this feature.


jakebeal:
Thank you; I will review and see if I can make use of the work that you have done.

PR #282, opened 2024-09-03
Implemented working prototype of 3.X sub_component <---> 2.X component conversion

### LINKED ISSUE #240: SBOL 2-3 converter - Sub_component (Priority 1)

Yehuda-Binik:
I implemented a first working version of `sub_component` <---> `component` conversion

A Unit test for a round trip starting from SBOL2 still needs to be written.

In addition, the prototype does not handle `hasLocation` and `orientation` properties. 

For the time being, I am conveerting all `sub_components` into 2.X `components`. However, this is not always the case, since "SBOL 2.x Component, Module, and FunctionalComponent objects map to SBOL 3.x SubComponent objects" as per the documentation. 

I don't necessarily feel comfortable merging yet, but I would appreciate feedback on what I have at the moment.


sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -142,38 +145,58 @@ def visit_combinatorial_derivation(self, a: sbol3.CombinatorialDerivation):
...
+        comp_def2.sequences = comp3.sequences
+        if comp3.features:
+            for feature in comp3.features:
+                if type(feature) == sbol3.subcomponent.SubComponent:
+                    self.visit_sub_component(feature, comp_def2)
+                elif type(feature) == sbol3.compref.ComponentReference:
+                    try:
+                        self.visit_component_reference(feature)
+                    except NotImplementedError as e:
+                        # highlights the error message in red.
```
jakebeal, line 170: I don't believe that you should be swallowing the error here and turning it into a print statement.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -142,38 +145,58 @@ def visit_combinatorial_derivation(self, a: sbol3.CombinatorialDerivation):
...
+                        print(f"\033[91m{e}\033[0m")
+                else:
+                    raise NotImplementedError(
+                        'Conversion of Component features from SBOL3 to SBOL2 not yet implemented')
```
jakebeal, line 174:
```suggestion
                        f'Conversion of Component feature of type {type(feature)} from SBOL3 to SBOL2 not yet implemented')
```

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -142,38 +145,58 @@ def visit_combinatorial_derivation(self, a: sbol3.CombinatorialDerivation):
...
+        if comp3.interactions:
+            for interaction in comp3.interactions:
+                try:
+                    self.visit_interaction(interaction)
+                except NotImplementedError as e:
+                    print(f"\033[91m{e}\033[0m")
```
jakebeal, line 180: Same as above - do not swallow errors here, or below

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -404,27 +436,56 @@ def visit_component_definition(self, cd2: sbol2.ComponentDefinition):
...
+        if comp2.roleIntegration:
+            sub3.role_integration = comp2.roleIntegration
+        if comp2.sourceLocations:
+            sub3.source_locations = comp2.sourceLocations
+        sub3.instance_of = comp2.definition
+        comp3.features += [sub3]
+        identity_mappings[sub3.identity] = comp2.identity
+
+    def handle_subcomponent_identity_triple_surgery(self, identity_mappings):
+        with tempfile.TemporaryDirectory() as tmpdir:
```
jakebeal, line 474: I don't think that you should be reading and writing tempfiles here. This could result in large numbers of complete read-write cycles in the course of a single conversion and be extremely slow.

I would suggesting saving up all of the identity surgery plans into a single shared datastructure, then doing the surgery on the graph like it's done over here: https://github.com/SynBioDex/SBOL-utilities/blob/9331964d61ff7f85ed91eee8e5f96cafb59665c5/sbol_utilities/conversion.py#L27

jakebeal (commented):
I've added my current thoughts.

You will also need to turn on GitHub actions in your branch and ensure that it passes when tested by GitHub

Issue #283, opened 2024-09-04
SBOL 2 to 3 Converter: Empty ComponentDefinition conversion


fxbuson:
I tried converting something very minimal with the python native converter:

```
import sbol2
from sbol_utilities import conversion

doc2 = sbol2.Document()
comp = sbol2.ComponentDefinition()
doc2.add(comp)

doc3 = conversion.convert2to3(doc2, use_native_converter = True)
```

(Note this converts with no problems in the JS converter)

The first error I get is related to namespaces. it seems like the current code does not accept None namespaces when it clearly should (there is even a default mode that should return None):
```
r._sbol3_namespace(self, obj2)
    351     return namespaces[0]
    352 # Check if the object starts with any of the provided namespaces
--> 353 for namespace in self.namespaces:
    354     if obj2.identity.startswith(namespace):
    355         return namespace

TypeError: 'NoneType' object is not iterable
```

I then tried giving it a namespace with:

`doc3 = conversion.convert2to3(doc2, namespaces = [sbol2.DEFAULT_NS], use_native_converter = True)`

I then got an error when extracting the identity to sbol3:

```
File ~\AppData\Local\Programs\Python\Python311\Lib\site-packages\sbol3\identified.py:60, in extract_display_id(identity)
     58 msg += ' A displayId MUST be composed of only alphanumeric'
     59 msg += ' or underscore characters and MUST NOT begin with a digit.'
---> 60 raise ValueError(msg)
```

We can see from the code in [pysbol3.identified.py](https://github.com/SynBioDex/pySBOL3/blob/main/sbol3/identified.py) that it tries to extract whatever is after the last forward slash:

```
    [...]
    display_id = parsed.path.split('/')[-1]
    if is_valid_display_id(display_id):
        return display_id
    else:
        msg = f'"{display_id}" is not a valid displayId.'
        msg += ' A displayId MUST be composed of only alphanumeric'
        msg += ' or underscore characters and MUST NOT begin with a digit.'
        raise ValueError(msg)
```

But the identity in SBOL2 ends with the version, not the ID. In this case my component had identity = 'http://examples.org/ComponentDefinition/example/1'. If I trim the identity first:

```
comp.identity = "/".join(comp.identity.split('/')[:-1])
```

Then the conversion runs ok. Should I just implement the trimming or is this expected to break something else? And for the namespace, should the function be looking for a None namespace or is this expected behaviour?

fxbuson:
[...]

PR #285, opened 2024-09-04
handle None namespace and version suffix in conversion

fxbuson:
As mentioned in Issue #283, the current version of the converter doesn't know how to handle identities that end in the version number. You get these in the case of using `sbol2.ConfigOptions.SBOL_COMPLIANT_URIS` as `True`, so I added a function `_sbol3_identity()` to be used whenever converting an identity value from sbol2 to sbol3, and exchanged each case of an sbol3 object being assigned an identity to use that function.

The converter also didn't include the version number in the identity when converting back to sbol2, so I included that in `_convert_identified()`. Note that this will default to '1', as per the SBOL 3.1.0 specification, section 10.2, item 3:

> When SBOL 3.x Identified object is converted to SBOL 2.x, if its identity is a URL, then the identity of 
the SBOL 2.x object SHOULD be constructed as [SBOL3 identity]/[SBOL2 version]. If the object does
not have an SBOL2 version property, then its version SHOULD default to 1. 

Another problem I ran into was with `None` namespaces when converting to sbol3, so I added that case to `_sbol3_namespace()`.

Now, my conversion of a simple document with an example ComponentDefinition should work. The roundtrip will enforce the use of the version at the end of the object's identity though, overriding the option `sbol2.ConfigOptions.SBOL_COMPLIANT_URIS` if it was set to `False`


PrashantVaidyanathan:
@jakebeal @tcmitchell - Just wanted to tag you in this. Are you happy with this?

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -73,6 +73,10 @@ def _convert_identified(self, obj3: sbol3.Identified, obj2: sbol2.Identified):
         obj2.description = self._value_or_property(obj3, obj3.description, 'http://purl.org/dc/terms/description')
         obj2.wasDerivedFrom = obj3.derived_from
         obj2.wasGeneratedBy = obj3.generated_by
+
+        # Add version to identity
+        obj2.identity = obj2.identity + "/" + obj3.sbol2_version
```
jakebeal: I don't think it's safe to assume that the `sbol2_version` property is necessarily set.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -341,6 +345,11 @@ def _convert_toplevel(self, obj2: sbol2.TopLevel, obj3: sbol3.TopLevel):
         self._convert_identified(obj2, obj3)
         obj3.attachments = [a.identity for a in obj2.attachments]
 
+    def _sbol3_identity(self, obj2: sbol2.Identified):
+        if obj2.identity.endswith("/" + obj2.version):
```
jakebeal:
```suggestion
        if obj2.version and obj2.identity.endswith("/" + obj2.version):
```
The `obj2.version` property is not required to be set. If this is the case, then this conversion might not work correctly.

jakebeal (changes requested):
This also needs to have the GitHub action tests run on it (turn them on in your fork) and needs to have new test cases added that are capable of reproducing the failures in #283 if these changes are not added.

Issue #310, opened 2025-04-15
Test files are not SBOL compliant


fxbuson:
According to the SBOL 2.3 specification, section 12.3: "Compliant SBOL Objects", item 6:

> If a compliant SBOL object has a version, then its identity property MUST contain a URI of the form “〈persistentIdentity〉/〈version〉”.

Right now, test files such as [BBa_J23101.xml](https://github.com/SynBioDex/SBOL-utilities/blob/develop/test/test_files/BBa_J23101.xml) have a version, but don't have the "/〈version〉" on their identity:

 ```xml
...
<sbol:ComponentDefinition rdf:about="https://synbiohub.org/public/igem/BBa_J23101">
  <ns0:mutableDescription>later</ns0:mutableDescription>
  <ns0:ownedBy rdf:resource="https://synbiohub.org/user/myers"/>
  <sbol:type rdf:resource="http://www.biopax.org/release/biopax-level3.owl#DnaRegion"/>
  <sbol:version>1</sbol:version>
...
```

If I create a file with pysbol2 with `SBOL_COMPLIANT_URIS` as `True`, the identity will have the "/〈version〉".

Similarly, exports from SynbioHub right now also include the version in the identity:

```xml
...
<sbol:ComponentDefinition rdf:about="https://synbiohub.org/public/igem/BBa_J23101/1">
  <sbol:persistentIdentity rdf:resource="https://synbiohub.org/public/igem/BBa_J23101"/>
  <sbol:displayId>BBa_J23101</sbol:displayId>
  <sbol:version>1</sbol:version>
...
```

I propose we change these test files to versions with the compliant format, adding their versions to the end of their Identities.

This is important because we are generating new test files for the new converter with pysbol2, and we need to be able to test the changes we make.

fxbuson:
Tagging @PrashantVaidyanathan @jakebeal @cjmyers to have a look at this

jakebeal:
I would say that it's important to keep these objects around as is: they are legal _non-compliant_ SBOL2. 

However, I think it would be OK to also add compliant variants of these objects, and to clearly mark which are and are not compliant and why in their comments / descriptions.

jakebeal:
I'm going to reopen this, since per @fxbuson 's comment on PR #315, these files will not yet be handled without errors.

To close this issue, update the converter so that these files are able to round-trip from SBOL2 to SBOL3 to SBOL2 without error and while still preserving their original (non-compliant) URI.

PR #311, opened 2025-04-16
Commented out tests with non-compliant files

fxbuson:
Since converter test files are not compliant (see issue #310) and there could be various reasons for non-compliance, I discussed with @PrashantVaidyanathan, @cjmyers and @jakebeal and found best to keep the files, but handle the conversion of them later.

For the efforts during Harmony 2025 we will only try to achieve conversions involving compliant SBOL2 files (taken from SBOL Test Suite or generated with pySBOL2)

Test files should be validated at https://validator.sbolstandard.org/ before being commited


fxbuson:
Will close this because of issues in test_calculate_complexity_scores.
I think this happens because my repository has different IDT credentials

PR #312, opened 2025-04-16
Commented out tests with non-compliant files

### LINKED ISSUE #310: Test files are not SBOL compliant

fxbuson:
Since converter test files are not compliant (see issue #310) and there could be various reasons for non-compliance, I discussed with @PrashantVaidyanathan, @cjmyers and @jakebeal and found best to keep the files, but handle the conversion of them later.

For the efforts during Harmony 2025 we will only try to achieve conversions involving compliant SBOL2 files (taken from SBOL Test Suite or generated with pySBOL2)

Test files should be validated at https://validator.sbolstandard.org/ before being commited

### DIFF
--- a/test/test_sbol2_sbol3_direct.py
+++ b/test/test_sbol2_sbol3_direct.py
@@ -15,6 +15,9 @@ TEST_FILES = Path(__file__).parent / 'test_files'
 class TestDirectSBOL2SBOL3Conversion(unittest.TestCase):
 
     # TODO: turn on validation
+
+    # J23101.xml is not SBOL compliant. Leaving conversions involving it for after the compliant converter is done
+    '''
     def test_3to2_conversion(self):
         """Test ability to convert a simple part from SBOL3 to SBOL2"""
         # Load an SBOL3 document and check its contents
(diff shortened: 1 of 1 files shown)


PrashantVaidyanathan (approved):

PrashantVaidyanathan:
Sounds good. We'll add the tests back in a separate PR - but this should unblock us for now.

PR #315, opened 2025-04-17
Conversion and test of an empty Component Definition

fxbuson:
Following from discussions on Harmony 2025 and issue #283, I made some changes to handle URIs in SBOL compliant files.

Main points:

- If an SBOL2 object has a version, we should move it to the middle of the URI in SBOL3, following the format \<namespace>/\<sbol2_version>/<collection + local>/\<displayId>
- Also, the object should have a backport#sbol2_version to store the information of the URI change
- On the conversion back from SBOL3, the "/\<version>" is stripped from the middle of the URI and added to the end
- This was packaged into two functions (_sbol2_identity() and _sbol3_identity), which is now referenced in the visitors for [SBOL3]Component and [SBOL2]Conponent Definition, respectively

This PR also handles a minor error that showed up when converting objects with `namespace` = `None`. I just check if there is a namespace list before going through it.

The conversion of Collection and Implementation is now outdated and untested, which will be recovered in a later PR.


sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -83,6 +83,12 @@ def _convert_identified(self, obj3: sbol3.Identified, obj2: sbol2.Identified):
         obj2.description = self._value_or_property(obj3, obj3.description, 'http://purl.org/dc/terms/description')
         obj2.wasDerivedFrom = obj3.derived_from
         obj2.wasGeneratedBy = obj3.generated_by
+
+        if obj2.version:
+            # TODO replace fragile string manipulation with robust path handling
```
jakebeal: I think this should be able to use the `_sbol2_identity` function. It would be good to ensure consistency. If it can't that's worrisome.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -95,10 +101,19 @@ def _convert_toplevel(self, obj3: sbol3.TopLevel, obj2: sbol2.TopLevel):
...
             obj.sbol2_version = sbol3.TextProperty(obj, BACKPORT2_VERSION, 0, 1)
-        # TODO: since version is optional, if it's missing, should this be returning '1' or None?
-        return obj.sbol2_version or '1'
+        return obj.sbol2_version or None
+
+    def _sbol2_identity(self, obj3: sbol3.Identified):
+        """Generate an SBOL2 identity for an SBOL3 object"""
+        identity = obj3.identity
+        if self._sbol2_version(obj3):
+            # TODO replace fragile string manipulation with robust path handling
```
jakebeal: As above: please link TODO to issue.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -160,7 +176,7 @@ def visit_component(self, cp3: sbol3.Component):
                     sbol3.SBO_NON_COVALENT_COMPLEX: sbol2.BIOPAX_COMPLEX}
         types2 = [type_map.get(t, t) for t in cp3.types]
         # Make the Component object and add it to the document
-        cp2 = sbol2.ComponentDefinition(cp3.identity, types2, version=self._sbol2_version(cp3))
+        cp2 = sbol2.ComponentDefinition(self._sbol2_identity(cp3), types2, version=self._sbol2_version(cp3))
```
jakebeal, line 179: This change needs to be applied everywhere, not just here.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -363,6 +379,26 @@ def _convert_toplevel(self, obj2: sbol2.TopLevel, obj3: sbol3.TopLevel):
...
+        # Getting only persistentIdentity will remove /<version> from the identity
+        identity = obj2.persistentIdentity
+
+        # Get namespace for sbol3 conversion
+        curr_namespace = parse_namespace(identity)
+        sbol3_namespace = self._sbol3_namespace(obj2)
+
+        # check for SBOL2 version and move it to middle of path
+        if obj2.version:
+            # TODO fix fragile string parsing with robust path handling
```
jakebeal: Ensure there is an open issue for this and name it in this TODO note.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -363,6 +379,26 @@ def _convert_toplevel(self, obj2: sbol2.TopLevel, obj3: sbol3.TopLevel):
         self._convert_identified(obj2, obj3)
         obj3.attachments = [a.identity for a in obj2.attachments]
 
+    def _sbol3_identity(self, obj2: sbol2.Identified):
+        """Generate an SBOL3 identity for an SBOL2 object"""
+
+        # Getting only persistentIdentity will remove /<version> from the identity
```
jakebeal, line 385: Will this work for non-compliant objects?

test/test_sbol2_sbol3_direct.py
```diff
@@ -149,5 +149,24 @@ def test_2to3_collection_conversion(self):
...
+            # Load an SBOL2 document and check its contents
+            doc2 = sbol2.Document()
+            doc2.read(TEST_FILES / 'empty_componentDefinition.xml')
+            # Convert to SBOL3 and check contents
+            doc3 = convert2to3(doc2, use_native_converter=True)
+            self.assertEqual(len(doc3.validate()), 0)
+            with tempfile.TemporaryDirectory() as tmpdir:
+                tmp3 = Path(tmpdir) / 'doc3.nt'
+                doc3.write(tmp3)
+                #self.assertFalse(file_diff(str(tmp3), str(TEST_FILES / 'BBa_J23101_patched.nt')))
```
jakebeal: This should be replaced with an SBOL3 serialization of the expected values for `empty_componentDefinition.nt`

jakebeal (changes requested):
Basic works looks solid; a few improvements are needed before merge.

sbol_utilities/sbol3_sbol2_conversion.py
fxbuson: The problem here is the `identity` should be versioned, but the `persistentIdentity` shouldn't.
Since we're using the configuration `SBOL_COMPLIANT_URIS` as `False` here, when we create the object the `identity` and `persistentIdentity` are the same. This needs to be patched after making the object, which is what I'm doing here.

sbol_utilities/sbol3_sbol2_conversion.py
fxbuson: As discussed I'll bring the changes from my other branch replacing the tests into here

sbol_utilities/sbol3_sbol2_conversion.py
fxbuson: If the object has a version and no versioning on the identity, this will then add a "/\<version>" on the identity.

Later on the script will try to strip whatever is after the last "/", so the identity will lose the displayId

So for now, it won't.

sbol_utilities/sbol3_sbol2_conversion.py
fxbuson: Done for all top level conversions in commit ca165c876ee8bf420f8d0266bdca5e09c3b96240
Child objects have their identities minted by pysbol3 and are currently being handled by the `update_identity` function.
I don't know if this is the best way to go, maybe we should just store the SBOL2 identity in a backport.

PrashantVaidyanathan (approved):
Looks good to me!

sbol_utilities/sbol3_sbol2_conversion.py
jakebeal: There's no requirement that SBOL3 objects have to use compliant IRIs, and it will be easier to track the relationship between the SBOL2 and SBOL3 object if the URIs are kept as similar as possible. As such, I recommend that we stick with the use of `update_identity` unless we discover a systematic problem with using SBOL2 identities.

sbol_utilities/sbol3_sbol2_conversion.py
jakebeal: That's OK for this PR, but I've re-opened #310 so that we don't lose track of this issue.

test/test_sbol2_sbol3_direct.py
```diff
@@ -149,5 +149,101 @@ def test_2to3_collection_conversion(self):
...
+            # Load an SBOL2 document and check its contents
+            doc2 = sbol2.Document()
+            doc2.read(TEST_FILES / 'minimal_sbol2_collection.xml')
+            # Convert to SBOL3 and check contents
+            doc3 = convert2to3(doc2, use_native_converter=True)
+            self.assertEqual(len(doc3.validate()), 0)
+            with tempfile.TemporaryDirectory() as tmpdir:
+                tmp3 = Path(tmpdir) / 'doc3.nt'
+                doc3.write(tmp3)
+                #self.assertFalse(file_diff(str(tmp3), str(TEST_FILES / 'BBa_J23101_patched.nt')))
```
jakebeal: Please add the equivalent SBOL3 `.nt` file to the tests and check it here. You can get it from an instance of `doc3.nt` produced by a test run.

test/test_sbol2_sbol3_direct.py
```diff
@@ -149,5 +149,101 @@ def test_2to3_collection_conversion(self):
...
+            # Load an SBOL2 document and check its contents
+            doc2 = sbol2.Document()
+            doc2.read(TEST_FILES / 'sbol_3to2_implementation_compliant.xml')
+            # Convert to SBOL3 and check contents
+            doc3 = convert2to3(doc2, use_native_converter=True)
+            self.assertEqual(len(doc3.validate()), 0)
+            with tempfile.TemporaryDirectory() as tmpdir:
+                tmp3 = Path(tmpdir) / 'doc3.nt'
+                doc3.write(tmp3)
+                #self.assertFalse(file_diff(str(tmp3), str(TEST_FILES / 'BBa_J23101_patched.nt')))
```
jakebeal: Please add the equivalent SBOL3 `.nt` file to the tests and check it here.

test/test_sbol2_sbol3_direct.py
```diff
@@ -149,5 +149,101 @@ def test_2to3_collection_conversion(self):
...
+            # Load an SBOL2 document and check its contents
+            doc2 = sbol2.Document()
+            doc2.read(TEST_FILES / 'seq_componentDefinition.xml')
+            # Convert to SBOL3 and check contents
+            doc3 = convert2to3(doc2, use_native_converter=True)
+            self.assertEqual(len(doc3.validate()), 0)
+            with tempfile.TemporaryDirectory() as tmpdir:
+                tmp3 = Path(tmpdir) / 'doc3.nt'
+                doc3.write(tmp3)
+                #self.assertFalse(file_diff(str(tmp3), str(TEST_FILES / 'BBa_J23101_patched.nt')))
```
jakebeal: Please add the equivalent SBOL3 `.nt` file to the tests and check it here.

jakebeal (changes requested):
Nice work and almost here - it just needs the three missing `.nt` files added for the tests and I think we're good to go.

jakebeal (approved):

PR #317, opened 2025-04-22
Refactor unittest.TestCase classes for the native converter

bbartley:
The roundtrip tests for the native converter re-use a lot of the same code.  Up til now, this code has been copy and pasted for each test case, and it's starting to get crufty.  So I factored out a more generalized method for handling conversion round trip tests.

Also, I added a final validation check for each round-tripped file.


test/test_sbol2_sbol3_direct.py
```diff
@@ -14,140 +14,84 @@
...
         # Convert to SBOL3 and check contents
         doc3 = convert2to3(doc2, use_native_converter=True)
         self.assertEqual(len(doc3.validate()), 0)
         with tempfile.TemporaryDirectory() as tmpdir:
             tmp3 = Path(tmpdir) / 'doc3.nt'
             doc3.write(tmp3)
-            self.assertFalse(file_diff(str(tmp3), str(TEST_FILES / 'BBa_J23101_patched.nt')))
+            self.assertFalse(file_diff(str(tmp3), str(TEST_FILES / rubric_filename)))
             doc2_loop = convert3to2(doc3, True)
-            # report = doc2.validate()
```
jakebeal, line 56: Please don't delete these two commented-out validation lines, or the TODO for turning them back on. We need to make sure the SBOL2 documents do validate someday.

test/test_sbol2_sbol3_direct.py
```diff
@@ -14,140 +14,84 @@
...
-            self.assertFalse(file_diff(str(tmp2), str(TEST_FILES / 'sbol_3to2_collection.xml')))
+            self.assertFalse(file_diff(str(tmp2), str(TEST_FILES / rubric_filename)))
             doc3_loop = convert2to3(doc2, use_native_converter=True)
             self.assertEqual(len(doc3_loop.validate()), 0)
             tmp3 = Path(tmpdir) / 'doc3_loop.nt'
             doc3_loop.write(tmp3)
-            self.assertFalse(file_diff(str(tmp3), str(TEST_FILES / 'sbol3_collection.nt')))
-    '''
+            self.assertFalse(file_diff(str(tmp3), str(TEST_FILES / test_filename)))
+            report = doc2.validate()
```
jakebeal: Why has the doc2 conversion moved down from above?
I think that it should be up when we generate the doc2 object, not after we write it.

jakebeal (changes requested):

test/test_sbol2_sbol3_direct.py
bbartley: I think I did some sloppy refactoring, cutting/pasting.  Anyway, https://github.com/SynBioDex/SBOL-utilities/pull/317/commits/bc085467746ce45e1e895c0321b80bd40fcebee7 now validates both the intermediate and the final round-tripped file generated by the converters.  

Also, added a few more comments just to make it a little more readable.

test/test_sbol2_sbol3_direct.py
bbartley: See above.

test/test_sbol2_sbol3_direct.py
```diff
@@ -27,18 +27,25 @@ def handle_2to3_conversion(self, test_filename, rubric_filename):
...
             tmp2 = Path(tmpdir) / 'doc2_loop.xml'
             doc2_loop.write(tmp2)
             self.assertFalse(file_diff(str(tmp2), str(TEST_FILES / test_filename)))
-            report = doc3.validate()
-            self.assertEqual(len(report), 0, f'Validation failed: {report}')
 
+    # Currently fails with the following validation error:
+    # sbol-12403: Strong Validation Error: The endedAtTime property of an Activity object is OPTIONAL and MAY contain
+    # a DateTime. Reference: SBOL Version 2.3.0 Section 12.8.1 on page 73 : https://synbiohub.org/public/igem/igem2sbol
+    @unittest.expectedFailure
```
jakebeal: Rather than turning off this test, I think that it would be good to check that precisely this error happens and no others.
And then add a TODO linked to an issue for resolving the validation error. If I remember correctly, I think this is actually a bug in the validator and not the converter.

jakebeal (changes requested):

test/test_sbol2_sbol3_direct.py
bbartley: This datetime value in the current test file (BBa_J23101.xml ) fails validation: 
`    <prov:endedAtTime>2017-03-06T15:00:00+00:00</prov:endedAtTime>`

However, a datetime that includes milliseconds _will_ validate.  (I figured this out by looking at test files in SBOLTestSuite that include a valid `endedAtTime`):
`    <prov:endedAtTime>2017-03-06T15:00:00.000+00:00</prov:endedAtTime>`

So, you may be correct that this is a problem with SBOLValidator, since these datetimes are semantically equivalent.

The most expedient fix is for me to just manually change the datetime in the test file to the accepted format.  Would that be acceptable?

jakebeal: That doesn't seem to have fixed it, given that it's still failing - I suspect that the conversion is putting it into that format. 
The options that I see are: 
1. Patching around that format issue
2. Catch the validation error and check if it's the expected issue

In either case, we need a follow-up issue filed as well.

bbartley: I attempted a patching solution (option 1) above but ran into challenges, so I'm working on implementing option 2 instead.  The big obstacle to patching is that the `DateTimePropertyMixin.from_user` `DateTimePropertyMixin.to_user` methods are not exposed in sbol3 (although, in sbol2, the analogous methods _are_ exposed and can be overridden).

I have opened an issue on SBOL-Validator to track this datetime issue:
https://github.com/SynBioDex/SBOL-Validator/issues/142

bbartley:
Replaced by #322

Issue #292, opened 2024-09-06
SBOL 2-3 converter - Sequence_constraint (Priority 2)


Yehuda-Binik:
# Progress Chart: 
| Property  | Test Written | Test Passed |
| :---: | :---: | :---: |
|  `subject`  |  |  |
|  `object`  |  |  |
|  `restriction`  |  |  |

PR #318, opened 2025-04-25
Convert SBOL2 Sequence Constraint Objects to SBOL3 Constraints Objects 

GeneCodeSavvy:
Resolving #292
I’ve fixed a few minor bugs and implemented the conversion method. For the `restriction` attribute of the `Constraint` object, for instance, if `http://sbols.org/v2#precedes` is used, should it be converted to `http://sbols.org/v3#precedes`?


GeneCodeSavvy:
@Gonza10V @jakebeal @PrashantVaidyanathan please take a look.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -637,8 +640,10 @@ def visit_sequence_annotation(self, sa2: sbol2.SequenceAnnotation):
         return f3, locations
  
     def visit_sequence_constraint(self, seq2: sbol2.sequenceconstraint.SequenceConstraint):
-        # Priority: 2
-        raise NotImplementedError('Conversion of SequenceConstraint from SBOL2 to SBOL3 not yet implemented')
+        subject = self.doc3.find(seq2.subject)
```
jakebeal: Calling `find` is very expensive. In this case, we know that the `subject` and `object` both have to be child objects of the same object that contains the SequenceConstraint, so we can search much more cheaply by just looking at the `Component` members of the parent.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -637,8 +640,10 @@ def visit_sequence_annotation(self, sa2: sbol2.SequenceAnnotation):
...
+        object = self.doc3.find(seq2.object)
+        restriction = seq2.restriction
```
jakebeal: The URIs need to move from the SBOL2 to SBOL3 namespace.
One good way to implement that is with a map like in the `visit_component` function: https://github.com/SynBioDex/SBOL-utilities/blob/6c44b64519851e180747fd18b4889e68a470f7b8/sbol_utilities/sbol3_sbol2_conversion.py#L156

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -637,8 +640,10 @@ def visit_sequence_annotation(self, sa2: sbol2.SequenceAnnotation):
...
+        return sbol3.Constraint(restriction, subject, object, name=seq2.name)
```
jakebeal: When you are mapping over the name like this, you are missing many of the other `Identified` object properties.
I recommend using `_convert_identified` following a similar pattern to here: https://github.com/SynBioDex/SBOL-utilities/blob/6c44b64519851e180747fd18b4889e68a470f7b8/sbol_utilities/sbol3_sbol2_conversion.py#L520

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -659,7 +664,7 @@ def convert3to2(doc3: sbol3.Document) -> sbol2.Document:
     return converter.doc2
 
 
-def convert2to3(doc2: sbol2.Document, namespaces=None) -> sbol3.Document:
+def convert2to3(doc2: sbol2.Document, namespaces=[]) -> sbol3.Document:
```
jakebeal: I do not think entry-point call should need to be changed in order to support your new code.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -459,10 +459,13 @@ def visit_component_definition(self, cd2: sbol2.ComponentDefinition):
...
-            raise NotImplementedError('Conversion of ComponentDefinition sequenceConstraints '
-                                      'from SBOL2 to SBOL3 not yet implemented')
+            for sc2 in cd2.sequenceConstraints:
+                sc3 = self.visit_sequence_constraint(sc2)
+                cp3.constraints.append(sc3)
+                self.update_identity(sc2, sc3)
+
         # Map over all other TopLevel properties and extensions not covered by the constructor
-        self._convert_toplevel(cd2, cp3)
+        self._convert_toplevel(cp3, cd2)
```
jakebeal: Why is this changing? I do not believe that it should be changed.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -89,7 +89,7 @@ def _convert_identified(self, obj3: sbol3.Identified, obj2: sbol2.Identified):
 
     def _convert_toplevel(self, obj3: sbol3.TopLevel, obj2: sbol2.TopLevel):
         """Map over the other properties of a TopLevel object"""
-        self._convert_identified(obj3, obj2)
+        self._convert_identified(obj2, obj3)
```
jakebeal: Why is this changing? I do not believe that it should be changed.

jakebeal (changes requested):
I have pointed out a few places that need improvement.

Please also add a test that exercises your code. You should also ensure that your tests run correctly before submitting (you can either ignore `TestIDTCalculateComplexityScore`, set it up to work on your fork, or we can set you up with a brach over in this repository)

sbol_utilities/sbol3_sbol2_conversion.py
GeneCodeSavvy: Noted. I have added a new parameter : `cp3` -> `sbol3.Component` object.

```python
def visit_sequence_constraint(self, seq2: sbol2.sequenceconstraint.SequenceConstraint, cp3: sbol3.Component):
        subject = cp3.find(seq2.subject)
```

sbol_utilities/sbol3_sbol2_conversion.py
GeneCodeSavvy: Thank you, added the following lines to address this..
```python
        constraint = sbol3.Constraint(restriction, subject, object, name=seq2.name)
        self._convert_identified(obj2=seq2, obj3=constraint)
        return constraint
```

sbol_utilities/sbol3_sbol2_conversion.py
GeneCodeSavvy: Sorry for not communicating properly in the commit description about this change. 

The root cause here was that `convert2to3` was using `namespaces=None` by default, so when it instantiates  

```python
SBOL2To3ConversionVisitor(doc2, namespaces=None)
```  

we end up with `self.namespaces is None`. Then, in  [`_sbol3_namespaces`](https://github.com/SynBioDex/SBOL-utilities/blob/6c44b64519851e180747fd18b4889e68a470f7b8/sbol_utilities/sbol3_sbol2_conversion.py#L366)

```python
def _sbol3_namespace(self, obj2):
    for namespace in self.namespaces:
        …
```

we get a `TypeError` because we can’t iterate over `None`. By switching the default in `convert2to3` to an empty list (`namespaces=[]`), we guarantee that `self.namespaces` is always iterable—just as the class’s own constructor default intended.  

This tweak only fixes that namespace‐handling bug; it doesn’t touch the SequenceConstraint→Constraint conversion logic at all.

GeneCodeSavvy:
> I have pointed out a few places that need improvement.
> 
> Please also add a test that exercises your code. You should also ensure that your tests run correctly before submitting (you can either ignore `TestIDTCalculateComplexityScore`, set it up to work on your fork, or we can set you up with a brach over in this repository)

I noticed the tests are currently failing on my end because I haven’t provided the IDT credentials yet. I’d prefer not to commit any secrets directly to the repository. Could you advise on the best way to supply those credentials securely?
If it’s easier, I’m also happy to open a separate pull request against any branch you suggest, or to follow whatever workflow you have in place for handling sensitive keys.

sbol_utilities/sbol3_sbol2_conversion.py
GeneCodeSavvy: I couldn't find any table specifying the conversion of Restriction URI from SBOL2 to SBOL3. I am assuming most of the URI remains same, just the #v2 -> #v3?

sbol_utilities/sbol3_sbol2_conversion.py
GeneCodeSavvy: I apologize. I mixed some functions in my test files. Undoing this.

sbol_utilities/sbol3_sbol2_conversion.py
GeneCodeSavvy: I apologize. I mixed some functions in my test files. Undoing this.

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: @GeneCodeSavvy I have also encountered the same problem while working on my own feature branch.  I started to implement the same fix as you, i.e., initialize the `namespaces` keyword argument as a list, but then remembered that using mutables (lists, dicts) as default arguments is considered an antipattern in Python.  See https://docs.python-guide.org/writing/gotchas/

So, I settled on the following solution:  https://github.com/SynBioDex/SBOL-utilities/commit/104535cafdabd20938ccd4028c5479c7eb6d5770

Unfortunately, as mentioned, my fix is still pending on a separate feature branch.  But I would recommend you go ahead and incorporate this fix into your branch.  Or find an alternative solution that doesn't involve using a mutable default argument (there are lots of docs/blogs/posts about how to approach this problem).

GeneCodeSavvy: @bbartley Thanks for the fix — I’ll incorporate it into my fork right away. Since I’m not working directly on a branch in this repo and will be working on the converter for a while, would you prefer that I open pull requests against your `feature/update-sbol2to3-converter` branch instead of `develop`?

sbol_utilities/sbol3_sbol2_conversion.py
jakebeal: Correct (at least for all of the ones that exist in v2). 

There are many in v3 that don't have equivalents in v2, so you can always go 2->3 but not always go 3->2. The ones that don't have a v2 equivalent should just stay with their original URI.

GeneCodeSavvy: Olay, for SBOL2 -> SBOL3, `sequenceConstraint.restriction.replace('/v2', '/v3')` should be enough ?

PR #322, opened 2025-05-22
 Refactor unittest.TestCase classes for the native converter

bbartley:
The roundtrip tests for the native converter re-use a lot of the same code. Up til now, this code has been copy and pasted for each test case and has gotten crufty. So I factored out a more generalized method for handling conversion round trip tests.

Also introduced three specific Exception classes to make debugging test failures easier

This replaces #317

### DIFF
--- a/test/test_sbol2_sbol3_direct.py
+++ b/test/test_sbol2_sbol3_direct.py
@@ -12,13 +12,21 @@ from sbol_utilities.sbol_diff import file_diff
 TEST_FILES = Path(__file__).parent / 'test_files'
 
 
-class SBOLValidationError(Exception): pass
+class SBOLValidationError(Exception):
+    """This exception class is intended for use within the context of TestCases to raise failures due to invalid SBOL
+    documents."""
+    pass
 
 
-class SBOL2to3ConversionError(Exception): pass
+class SBOL2to3ConversionError(Exception):
+    """This exception class is intended for use within the context of TestCases to raise failures converting SBOL2
+    documents to SBOL3."""
+    pass
 
-
-class SBOL3to2ConversionError(Exception): pass
+class SBOL3to2ConversionError(Exception):
+    """This exception class is intended for use within the context of TestCases to raise failures converting SBOL3
+    documents to SBOL2."""
+    pass
 
 
 class TestDirectSBOL2SBOL3Conversion(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


test/test_sbol2_sbol3_direct.py
```diff
@@ -12,6 +12,15 @@
 TEST_FILES = Path(__file__).parent / 'test_files'
 
 
+class SBOLValidationError(Exception): pass
```
jakebeal: I'd suggest adding a comment about how this is distinct from `sbol3.ValidationError` and `sbol2.SBOLError`.

test/test_sbol2_sbol3_direct.py
```diff
@@ -244,6 +253,82 @@ def test_seq_componentDefinition(self):
                 tmp2 = Path(tmpdir) / 'doc2_loop.xml'
                 doc2_loop.write(tmp2)
                 self.assertFalse(file_diff(str(tmp2), str(TEST_FILES / 'seq_componentDefinition.xml')))
+
+    def handle_2to3_conversion(self, test_filename, rubric_filename):
+        """Provides a re-usable test handler for converting SBOL2 to SBOL3 with different test files
+        test_filename: Name of an SBOL2 test file
```
jakebeal: Suggest changing the parameter declarations in the docstring to standard format (both here and elsewhere), so they'll link well in IDEs:
```suggestion
        :param test_filename: Name of an SBOL2 test file
```

test/test_sbol2_sbol3_direct.py
```diff
@@ -244,6 +253,82 @@ def test_seq_componentDefinition(self):
                 tmp2 = Path(tmpdir) / 'doc2_loop.xml'
                 doc2_loop.write(tmp2)
                 self.assertFalse(file_diff(str(tmp2), str(TEST_FILES / 'seq_componentDefinition.xml')))
+
+    def handle_2to3_conversion(self, test_filename, rubric_filename):
```
jakebeal: Suggest adding typing to function parameters, both here and elsewhere:
```suggestion
    def handle_2to3_conversion(self, test_filename: str, rubric_filename: str):
```

test/test_sbol2_sbol3_direct.py
```diff
@@ -244,6 +253,82 @@ def test_seq_componentDefinition(self):
...
+        # Convert to SBOL3 and check contents
+        doc3 = convert2to3(doc2, use_native_converter=True)
+        report = doc3.validate()
+        if len(report):
+            raise SBOLValidationError(report)
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmp3 = Path(tmpdir) / 'doc3.nt'
+            doc3.write(tmp3)
+            if file_diff(str(tmp3), str(TEST_FILES / rubric_filename)):
```
jakebeal: Calling this `rubric_filename` seems odd to me. Maybe `expected_converted_file`?

jakebeal (approved):
I've made a few suggestions for style improvements. Take them or leave them as you wish, but this is good to go from a functionality standpoint - please merge when you are ready.

Issue #293, opened 2024-09-06
SBOL 2-3 converter - Module_definition (Priority 3)


Yehuda-Binik:
# Progress Chart: 
| Property  | Test Written | Test Passed |
| :---: | :---: | :---: |
|  `roles`  |  |  |
|  `modules`  |  |  |
|  `functional_components`  |  |  |
|  `interactions`  |  |  |
|  `models`  |  |  |

bbartley:
Closed by #327

Issue #326, opened 2025-06-12
SBOL 2->3 converter: post-creation fix-up of links


bbartley:
The current approach for converting SBOL2 -> SBOL3 compliant URIs results in broken links.

For example, here is a broken link in this intermediate test file currently being used as the expected SBOL3:
https://github.com/SynBioDex/SBOL-utilities/blob/develop/test/test_files/sbol_3to2_implementation_compliant.nt#L6

Perhaps an important design decision here is:
Do we update all SBOL2 URIs that reference objects either internal or external to the Document (probably using some kind of regex pattern matching)?
or
Do we update only SBOL2 URIs that reference other objects internal the Document (probably using some kind of lookup table)?

jakebeal:
We should only ever update URIs that reference objects internal to the Document.
Objects outside of the document, by definition we cannot know if it's been converted.

jakebeal:
The plan that @bbartley has for tackling this is to do triple-surgery with the aid of a hashtable at the end of conversion. Every object of a triple will be checked against the hash table for mapping to a new identity.

The only case in which this will produce a behavior other than an identity update per the description above, is the case where somebody has abused a text field to fill it with a URI that matches the identity of an object in the document.
 
If that is the case, the text field will also be updated, and we are considering that to be a feature rather than a bug.

Issue #246, opened 2024-04-08
SBOL 2-3 converter - Component (Priority 2)


PrashantVaidyanathan:

Yehuda-Binik:
@jakebeal 

Me and @PrashantVaidyanathan were discussing this issue. It appears that `visit_component` has not yet been completed. Currently, the function implements correctly the following line on page 51 from the SBOL3 documentation: "SBOL 2.x `ComponentDefinition` objects map to SBOL 3.x `Component` objects. The type property is mapped according to Table 19. " However, the function does not implement the following line in the documentation: "SBOL 2.x `ModuleDefinition` objects map to SBOL 3.x `Component` objects with a type of SBO:0000241 (func- 18tional entity)." How should the function be modified to take this into account?

jakebeal:
Yes, it has not yet been implemented and needs to be. Not sure how much more to say than that...

bbartley:
An SBOL3 Component may actually map into SBOL2 as _both_ a `ComponentDefinition` and a `ModuleDefinition`.  Currently the `SBOL3To2ConversionVisitor.visit_component` method only converts to one or the other.  Leaving a note here that:

1. If there are any `interactions`, `interface`, or `models`, or SBO function entity type, then there must by a `ModuleDefinition`
2. If there are any `constraints`, `sequences`, or `features` with a `Location`, then there must be a `ComponentDefinition`.

PR #327, opened 2025-06-15
Feature/293 convert moduledefinition

bbartley:
Roundtrip conversion of ModuleDefinition to Component and back again (#293 )

Current test case focuses on a simple ModuleDefinition.  Future PRs will expand test coverage to child classes, e.g., Interaction, FunctionalComponent, etc.

Test files were created synthetically using a pySBOL script (as opposed to representing an SBOL file found "in the wild").  Question: Do we want to check in this script? (It may be helpful for building out other representations to use as test cases).

While working on this, I noticed there's an annoying workaround that is required for our current round-trip tests to work.  In order for the triples in the original file to exactly match the triples in the round-tripped file, we have to artificially introduce a triple with `backport:sbol3namespace`  in the test file (see https://github.com/SynBioDex/SBOL-utilities/blob/8645fbe33b46173048dcfbb3300ff4bdfbaa8330/test/test_files/sbol_3to2_moduledefinition.xml#L4). 

(See https://github.com/SynBioDex/SBOL-utilities/issues/356 and https://github.com/SynBioDex/SBOL-utilities/issues/357)

### DIFF
--- a/sbol_utilities/sbol3_sbol2_conversion.py
+++ b/sbol_utilities/sbol3_sbol2_conversion.py
@@ -175,24 +175,36 @@ class SBOL3To2ConversionVisitor:
                     sbol3.SBO_SIMPLE_CHEMICAL: sbol2.BIOPAX_SMALL_MOLECULE,
                     sbol3.SBO_NON_COVALENT_COMPLEX: sbol2.BIOPAX_COMPLEX}
         types2 = [type_map.get(t, t) for t in cp3.types]
-        # Make the Component object and add it to the document
-        cp2 = sbol2.ComponentDefinition(self._sbol2_identity(cp3), types2, version=self._sbol2_version(cp3))
-        self.doc2.addComponentDefinition(cp2)
-        # Convert the Component properties not covered by the constructor
-        cp2.roles = cp3.roles
-        cp2.sequences = cp3.sequences
-        if cp3.features:
-            raise NotImplementedError('Conversion of Component features from SBOL3 to SBOL2 not yet implemented')
-        if cp3.interactions:
-            raise NotImplementedError('Conversion of Component interactions from SBOL3 to SBOL2 not yet implemented')
-        if cp3.constraints:
-            raise NotImplementedError('Conversion of Component constraints from SBOL3 to SBOL2 not yet implemented')
-        if cp3.interface:
-            raise NotImplementedError('Conversion of Component interface from SBOL3 to SBOL2 not yet implemented')
-        if cp3.models:
-            raise NotImplementedError('Conversion of Component models from SBOL3 to SBOL2 not yet implemented')
-        # Map over all other TopLevel properties and extensions not covered by the constructor
-        self._convert_toplevel(cp3, cp2)
+
+        # Determine whether Component maps to a ComponentDefinition or ModuleDefinition
+        if sbol3.SBO_FUNCTIONAL_ENTITY not in cp3.types:
+            # Make the Component object and add it to the document
+            cp2 = sbol2.ComponentDefinition(self._sbol2_identity(cp3), types2, version=self._sbol2_version(cp3))
+
+            # Convert the Component properties not covered by the constructor
+            cp2.roles = cp3.roles
+            cp2.sequences = cp3.sequences
+            for f in cp3.features:
+                raise NotImplementedError('Conversion of Component features from SBOL3 to SBOL2 not yet implemented')
+            if cp3.interactions:
+                raise NotImplementedError('Conversion of Component interactions from SBOL3 to SBOL2 not yet implemented')
+            if cp3.constraints:
+                raise NotImplementedError('Conversion of Component constraints from SBOL3 to SBOL2 not yet implemented')
+            if cp3.interface:
+                raise NotImplementedError('Conversion of Component interface from SBOL3 to SBOL2 not yet implemented')
+            if cp3.models:
+                raise NotImplementedError('Conversion of Component models from SBOL3 to SBOL2 not yet implemented')
+            self.doc2.addComponentDefinition(cp2)
+            self._convert_toplevel(cp3, cp2)
+        else:
+            # Component maps to an SBOL2 ModuleDefinition
+            mdef2 = sbol2.ModuleDefinition(self._sbol2_identity(cp3),
+                                           version=self._sbol2_version(cp3))
+            mdef2.roles = cp3.roles
+            mdef2.models = cp3.models
+            self.doc2.addComponentDefinition(mdef2)
+            self._convert_toplevel(cp3, mdef2)
+
 
     def visit_component_reference(self, a: sbol3.ComponentReference):
         # Priority: 3
@@ -584,8 +596,8 @@ class SBOL2To3ConversionVisitor:
         raise NotImplementedError('Conversion of Module from SBOL2 to SBOL3 not yet implemented')
 
     def visit_module_definition(self, md: sbol2.ModuleDefinition):
-        # Make the Component object and add it to the document
-        c3 = sbol3.Component(self._sbol3_identity(md), types=md.type, roles=md.roles, namespace=self._sbol3_namespace(md))
+        # ModuleDefinitions convert to SBOL3 Components annotated as "functional entities"
+        c3 = sbol3.Component(self._sbol3_identity(md), types=[sbol3.SBO_FUNCTIONAL_ENTITY], roles=md.roles, namespace=self._sbol3_namespace(md))
 
         for i2 in md.interactions:
             i3 = self.visit_interaction(i2)
--- a/sbol_utilities/sbol3_sbol2_conversion.py
+++ b/sbol_utilities/sbol3_sbol2_conversion.py
@@ -617,6 +617,7 @@ class SBOL2To3ConversionVisitor:
                 if fc.direction == 'http://sbols.org/v2#none':
                     c3.interface.nondirectionals.append(sc)
         self.doc3.add(c3)
+        self._convert_toplevel(md, c3)
 
     def visit_participation(self, p2: sbol2.Participation):
         p3 = sbol3.Participation(p2.roles, strip_sbol2_version(p2.participant))
--- a/sbol_utilities/sbol3_sbol2_conversion.py
+++ b/sbol_utilities/sbol3_sbol2_conversion.py
@@ -604,17 +604,18 @@ class SBOL2To3ConversionVisitor:
             c3.interactions.append(i3)
             self.update_identity(i2, i3)
 
-        c3.interface = sbol3.Interface()
-        for fc in md.functionalComponents:
-            sc = self.visit_functional_component(fc)
-            c3.features.append(sc)
-            self.update_identity(fc, sc)
-            if fc.direction == 'http://sbols.org/v2#in' or fc.direction == 'http://sbols.org/v2#inout':
-                c3.interface.inputs.append(sc)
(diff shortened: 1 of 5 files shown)


sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -175,24 +175,36 @@ def visit_component(self, cp3: sbol3.Component):
...
+            if cp3.models:
+                raise NotImplementedError('Conversion of Component models from SBOL3 to SBOL2 not yet implemented')
+            self.doc2.addComponentDefinition(cp2)
+            self._convert_toplevel(cp3, cp2)
+        else:
+            # Component maps to an SBOL2 ModuleDefinition
+            mdef2 = sbol2.ModuleDefinition(self._sbol2_identity(cp3),
+                                           version=self._sbol2_version(cp3))
+            mdef2.roles = cp3.roles
+            mdef2.models = cp3.models
```
jakebeal: I don't believe that this is safe: the models need to be converted. Replace this with a `NotImplementedError` like above?

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -584,26 +596,28 @@ def visit_module(self, a: sbol2.Module):
...
-            c3.features.append(sc)
-            self.update_identity(fc, sc)
-            if fc.direction == 'http://sbols.org/v2#in' or fc.direction == 'http://sbols.org/v2#inout':
-                c3.interface.inputs.append(sc)
-            if fc.direction == 'http://sbols.org/v2#out' or fc.direction == 'http://sbols.org/v2#inout':
-                c3.interface.outputs.append(sc)
-            if fc.direction == 'http://sbols.org/v2#none':
-                c3.interface.nondirectionals.append(sc)
+        if md.functionalComponents:
+            c3.interface = sbol3.Interface()
```
jakebeal: Interface shouldn't be unconditionally created - it should only be added if there is at least one FunctionalComponent with a direction other than `none`.

test/test_sbol2_sbol3_direct.py
```diff
@@ -299,6 +299,9 @@ def handle_2to3_conversion(self, test_filename: str, comparison_filename: str):
     def test_implementation_conversion(self):
         self.handle_2to3_conversion('sbol_3to2_implementation_compliant.xml', 'sbol_3to2_implementation_compliant.nt')
 
+    def test_moduledefinition_conversion(self):
+        self.handle_2to3_conversion('sbol_3to2_moduledefinition.xml', 'sbol_3to2_moduledefinition.nt')
```
jakebeal, line 304: Let us also test with `handle_3to2_conversion`, in the other half of the file.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -584,26 +596,28 @@ def visit_module(self, a: sbol2.Module):
         raise NotImplementedError('Conversion of Module from SBOL2 to SBOL3 not yet implemented')
 
     def visit_module_definition(self, md: sbol2.ModuleDefinition):
-        # Make the Component object and add it to the document
-        c3 = sbol3.Component(self._sbol3_identity(md), types=md.type, roles=md.roles, namespace=self._sbol3_namespace(md))
```
jakebeal, line 591: This looks like it would never have worked... was this actually tested before?

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -584,26 +596,28 @@ def visit_module(self, a: sbol2.Module):
...
-            sc = self.visit_functional_component(fc)
-            c3.features.append(sc)
-            self.update_identity(fc, sc)
-            if fc.direction == 'http://sbols.org/v2#in' or fc.direction == 'http://sbols.org/v2#inout':
-                c3.interface.inputs.append(sc)
-            if fc.direction == 'http://sbols.org/v2#out' or fc.direction == 'http://sbols.org/v2#inout':
-                c3.interface.outputs.append(sc)
-            if fc.direction == 'http://sbols.org/v2#none':
-                c3.interface.nondirectionals.append(sc)
+        if md.functionalComponents:
```
jakebeal, line 610: Is this change to avoid creating the Interface if there aren't functional components? If so, this isn't actually the way to go about it, because the decision should be based on direction markings, not on whether there are components.

sbol_utilities/sbol3_sbol2_conversion.py
```diff
@@ -175,24 +175,36 @@ def visit_component(self, cp3: sbol3.Component):
...
-            raise NotImplementedError('Conversion of Component constraints from SBOL3 to SBOL2 not yet implemented')
-        if cp3.interface:
-            raise NotImplementedError('Conversion of Component interface from SBOL3 to SBOL2 not yet implemented')
-        if cp3.models:
-            raise NotImplementedError('Conversion of Component models from SBOL3 to SBOL2 not yet implemented')
-        # Map over all other TopLevel properties and extensions not covered by the constructor
-        self._convert_toplevel(cp3, cp2)
+
+        # Determine whether Component maps to a ComponentDefinition or ModuleDefinition
+        if sbol3.SBO_FUNCTIONAL_ENTITY not in cp3.types:
```
jakebeal, line 185: We need to take more than the type into account in making this decision. An SBOL3 `Component` may actually map into SBOL2 as _both_ a ComponentDefinition and a ModuleDefinition. 

- If there are any `interactions`, `interface`, or `models`, or SBO function entity type, then there must by a `ModuleDefinition`
- If there are any `constraints`, `sequences`, or features with a `Location`, then there must be a `ComponentDefinition`.

jakebeal (changes requested):

jakebeal:
Regarding the question of the `backport:sbol3namespace` property: round-tripping starting from SBOL2 may end up with something different than round-tripping starting from SBOL3. I suspect you are having a difficulty because you created your SBOL2 and SBOL3 objects independently. I would suggest that we need two sets of files here: one where it originates in SBOL2 and one where it originates in SBOL3.

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: Recall that the `models` property is an associative property rather than compositional.  This is just copying the references.  For now, the URIs for the referenced models will remain in SBOL2 format and are not converted to SBOL3 compliant URIs, but that is an issue (#326) that is expected to be resolved later.

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: Nope, not tested!  This is an artifact of some code that got merged prematurely while still a WIP into develop during HARMONY.   This PR is building off of, cleaning up, and validating some of that initial work

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: To keep this PR from ballooning and maintain issue scope to the linked issue (#293 ) which deals with converting a ModuleDefinition, I would suggest maybe inserting a TODO and a note in (#246 ) which concerns converting a Component.  Does that sound reasonable?

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: Nice catch!

sbol_utilities/sbol3_sbol2_conversion.py
jakebeal: That's why I would tend to leave it as "NotImplementedError" for now (like it is above) rather than copy unconverted materials, which will cause errors that are likely to be harder for a user to interpret.

sbol_utilities/sbol3_sbol2_conversion.py
jakebeal: I'm OK with inserting a TODO that explains this.

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: On second thought, before making any further changes here, we need to clarify when it is appropriate to use the `Interface.nondirectionals` property.  Two issues:

1. If a `FunctionalComponent` has no direction,  in what cases, if any, would it be appropriate to add it to an `Interface` with a nondirectional `Feature`?
2. If a `FunctionalComponent` has both in and out directionality, would we add it to an `Interface` both as an input and output, or do we add it as a nondirectional `Feature`?

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: Added a TODO in cb84b740c2074379d20e89c3b3f89b0236a163f5 and left a note in #246

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: Unfortunately, until #326 is resolved, errors from unconverted URIs are inevitable.  Currently we initialize some reference properties by construction, e.g., `ComponentDefinition.sequences` and `Implementation.built`.   Since these properties are initialized by the constructor and they are required, there's no way to raise a `NotImplementedError`.  So, you're stuck with an unconverted URI.   I'd rather just go ahead and implement the copy now rather then leave it for later.

jakebeal: OK, just be sure to leave a TODO pointing to #326 with this information, so that we'll know how to deal with the problem in the future.

sbol_utilities/sbol3_sbol2_conversion.py
jakebeal: In the SBOL3 specification Section 10.2 on conversion between SBOL2 and SBOL3, there is the following:

> Every FunctionalComponent in an SBOL 2.x ModuleDefinition with a "direction" property that is not
"none" is listed in the Interface of its SBOL 3.x Component. The mapping from direction to interface properties is: "in"–>"inputs", "out"–>"outputs", "inout" –> "nondirectional". Finally, every Component with "access"="public" and "direction"="none" is listed as "nondirectional" in the Interface.

So:
1. If there is `direction`=`none`, then it should be `nondirectional` if and only if it has `access`=`public`
2. If it has `direction`=`inout`, then it should be placed in `nondirectional` (direction is a singleton property, so it cannot have both `input` and `output`.

Note that since two different things are mapping into `nondirectional`, then we will need to leave a reverse translation hint. I would suggest that for 3->2 the default case should be `nondirectional` -> `inout`, such that a hint only needs to be left behind for `direction`=`none` + `access`=`public`, which is case that was not intended but sometimes used.

bbartley: To prevent further sprawl in this PR, I have deferred further implementation of `FunctionalComponent` to https://github.com/SynBioDex/SBOL-utilities/pull/355 -- I have copied these notes over.

test/test_sbol2_sbol3_direct.py
bbartley: Added in 883d7268f82323c4e4f4b8d98e386f87e7bebf8a

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: For now I have marked creation of `Interface` as not implemented, and deferred it to a different PR: https://github.com/SynBioDex/SBOL-utilities/pull/355

sbol_utilities/sbol3_sbol2_conversion.py
bbartley: TODO added in eaa1a75052c35699571126dd1bbc5231c38923d9

jakebeal (approved):

PR #355, opened 2025-07-15
Feature/295 convert functional component

bbartley:
Per review comments in https://github.com/SynBioDex/SBOL-utilities/pull/327

1.  If a `FunctionalComponent` has `direction` == `none`, then it should be added to an `Interface` as a `nondirectional` if and only if it has `access` == `public`
2.  If a `FunctionalComponent` it has `direction` = `inout`, then it should be placed in an `Interface` as a  `nondirectional` rather than adding it to both an `Interface's` `inputs` and `outputs`

Note that since two different things are mapping into `nondirectional`, then we will need to leave a reverse translation hint. I would suggest that for 3->2 the default case should be `nondirectional` -> `inout`, such that a hint only needs to be left behind for` direction` = `none` + `access` = `public`, which is case that was not intended but sometimes used.


jakebeal (approved):
Approved, on the assumption that conflicts will be resolved without changes to functionality.

Issue #357, opened 2025-07-15
Backport annotations cause gotcha when diffing round-tripped files


bbartley:
Our current mechanism for converter tests involves testing the original file with the round-tripped file for an exact match of triples.  This creates a complication, since round-tripping adds backport annotations– thus, the original file and its round-trip never exactly match.   To get around this, we have to artificially add backport annotations to the original file to ensure it matches the round-tripped file.  This is a subtle and annoying gotcha when trying to generate test files for new converter tests. 

At the moment, I'm uncertain what is the best solution, but likely it involves some special manipulation of backport annotations by the `handle_3to2_conversion` and `handle_2to3_conversion` methods.  And the solution in each of these methods will need to be slightly different, since round-tripping starting from SBOL2 may end up with something different than round-tripping starting from SBOL3.

jakebeal:
One good way to do this would be to add a "compare, ignoring backport annotations" method, which would strip out the backport annotation triples before serializing and comparing.

That will mean that we don't have to modify the original files, and we should always be safe to ignore backport materials, because the only things that will use them are the converters, for which the backport annotation integrity is being implicitly tested by the round-trip.

jakebeal:
Note that a good source of triple manipulation code that can be adapted for this is the non-native "javascript wrapper" converters, which have to do a bunch of "triple grep" operations in order to deal with known bugs in the non-native converter.

PR #359, opened 2025-07-24
Ignore Backport Properties in SBOL Comparison

GeneCodeSavvy:
While writing tests for **Attachment** objects, I encountered a bug related to namespace properties during SBOL document conversions:

* Converting an **SBOL2** document to **SBOL3** adds a new `namespace` property.
* When converting the resulting **SBOL3** document back to **SBOL2**, a **backport namespace** property is added.
* This causes test failures because `sbol_diff.file_diff()` performs a strict one-to-one file comparison, and the added backport property results in mismatches.

This PR ensures that backport properties are ignored during file comparisons, preventing false test failures.

Related to #357


jakebeal (changes requested):
I very much like this and think that it is a very useful contribution. 

I would like to suggest adding one additional tweak, however: let the function take an argument to say whether it is stripping out SBOL2 or SBOL3 backport properties, and then strip out only those.

The reason for this is to make sure that backport properties are properly cleaned up after a round trip. If we have an SBOL2 file, then we want to strip SBOL3 backport properties, but should flag a problem if there is an SBOL2 backport property still in the file. Likewise, an SBOL3 file may have SBOL2 backport properties, but should not have any SBOL3 backport properties.

Gonza10V:
Agree with Jake, seems reasonable and is not difficult to implement. After this change we should be able to detect if the objects have unintended backport properties.

GeneCodeSavvy:
I have added functions to detect the sbol version from the files, and error out if the the document contains invalid backport properties. Therefore,  Catching mistakes in the convertor

bbartley:
@GeneCodeSavvy  @Gonza10V  Correct me if I'm wrong, but does this solution permanently alter the files being diffed by stripping them of their backport annotations?  Wouldn't we want to restore those triples after performing the diff?

Gonza10V:
@bbartley Hmm, I think we dont use those files afterwards, the strip should be just for tests. If an user would like to strip backport properties and then use the file maybe should make a copy of the file and process one and then pass the other.

bbartley:
@Gonza10V I'm concerned because `sbol_diff.file_diff()` is part of a general utility package, and it's entirely possible that other users might use it for different purposes than our round-trip tests.  Also, a `diff` operation that modifies its input files is very unusual, and could lead to some nasty surprises for unwitting users.  So, I would suggest _not_ making this the default behavior.  Rather, we could add an optional flag such as `strip_backport_annotations=False` that preserves the default behavior, but then have the round-trip tests use `strip_backport_annotations=True`

Gonza10V:
I like the proposal. The diff function should not change the files that is comparing. The function should have `strip_backport_annotations` as parameter, and be False by default so users are aware of the stripping. Let me know your thoughts.

GeneCodeSavvy:
No function created in this PR affect the original files. The Graph object created from the files are being manipulated, in memory and are not saved.
cc @Gonza10V @bbartley

bbartley:
@GeneCodeSavvy sorry, you are right that _the files are not changed_ -- rather I should have said _the Graphs are changed_ .  But the same point applies -- a diff operation that modifies the comparands is unusual and unexpected, so we need some safeguard so that users are not caught unaware.  Also, there may be use cases where a user wants to perform a diff and then continue to use the Graphs unmodified afterwards.

GeneCodeSavvy:
Thank you pointing that out @bbartley. I have changed the function, so that graphs are also not changed. I have also added a new flag `strip_backport_properties` that is false by default, so it should not break previous uses and users have to consciously set the flag to true to ignore the backport properties while diffing.

Gonza10V (approved):

Gonza10V:
@jakebeal  can you take a look to this? this looks good to me and you can merge if it looks good to you.

sbol_utilities/sbol_diff.py
```diff
@@ -16,9 +18,120 @@ def _load_rdf(fpath: Union[str, bytes, os.PathLike]) -> rdflib.Graph:
...
+        if str(namespace) == sbol2_namespace:
+            return 'sbol2'
+        elif str(namespace) == sbol3_namespace:
+            return 'sbol3'
+
+    # If no namespace bindings found, check for SBOL URIs in the actual triples
+    for _, p, _ in graph:
+        # Check predicates for SBOL namespace URIs
+        predicate_str = str(p)
+        if sbol2_namespace in predicate_str:
```
jakebeal: These should be testing the prefix with `startswith` rather than `in`.
It's unlikely to go wrong given the constraints on URIs, but it's good code hygiene.

sbol_utilities/sbol_diff.py
```diff
@@ -16,9 +18,120 @@ def _load_rdf(fpath: Union[str, bytes, os.PathLike]) -> rdflib.Graph:
...
+
+    :param graph: the RDF graph to analyze
+    :return: 'sbol2', 'sbol3', or None if version cannot be determined
+    """
+    # Look for SBOL namespace declarations
+    sbol2_namespace = 'http://sbols.org/v2#'
+    sbol3_namespace = 'http://sbols.org/v3#'
+
+    # Check bound namespaces first
+    for _, namespace in graph.namespaces():
```
jakebeal: If a document contains both sbol2 and sbol3 namespaces, this iteration will produce unpredictable results, depending on the order of iteration through the graph. Instead of an iteration, use a set comprehension to collect all the namespaces. Then if both are present, return `None`, and if only one is present return that one.

The same change should be made to search of predicates below.

sbol_utilities/sbol_diff.py
```diff
@@ -16,9 +18,120 @@ def _load_rdf(fpath: Union[str, bytes, os.PathLike]) -> rdflib.Graph:
...
+
+    for s, p, o in graph:
+        predicate_str = str(p)
+        if not predicate_str.startswith(BACKPORT_NAMESPACE):
+            clean_graph.add((s, p, o))
+
+    return clean_graph
+
+
+def _diff_graphs(g1: rdflib.Graph, g2: rdflib.Graph,
```
jakebeal, line 116: Function needs a docstring

sbol_utilities/sbol_diff.py
```diff
@@ -16,9 +18,120 @@ def _load_rdf(fpath: Union[str, bytes, os.PathLike]) -> rdflib.Graph:
...
+                 strip_backport_properties: bool = False) -> Tuple[rdflib.Graph, rdflib.Graph, rdflib.Graph]:
+    # Remove backport properties from both graphs before comparison
+    g1_sbol_version = _detect_sbol_version(g1)
+    g2_sbol_version = _detect_sbol_version(g2)
+
+    if g1_sbol_version != g2_sbol_version:
+        raise TypeError('SBOL documents are of different versions, cannot compare')
```
jakebeal: I see no reason that we can't compare documents of different versions - they just won't be equal!

sbol_utilities/sbol_diff.py
```diff
@@ -16,9 +18,120 @@ def _load_rdf(fpath: Union[str, bytes, os.PathLike]) -> rdflib.Graph:
...
+
+    if g1_sbol_version:
+        validate_backport_properties(g1, g1_sbol_version)
```
jakebeal: Why are we validating in a comparison function? This feels inappropriate.
In test functions, we should be comparing separately from validating backport properties.

sbol_utilities/sbol_diff.py
```diff
@@ -16,9 +18,120 @@ def _load_rdf(fpath: Union[str, bytes, os.PathLike]) -> rdflib.Graph:
...
+    if g2_sbol_version:
+        validate_backport_properties(g2, g2_sbol_version)
+
+    g1_to_compare = g1
```
jakebeal, line 134: I think that everything above this point is not needed in this function.

test/test_sbol_diff.py
```diff
@@ -51,6 +63,129 @@ def test_doc_diff(self):
...
+        version2 = sbol_utilities.sbol_diff._detect_sbol_version(graph2_xml)
+        self.assertEqual('sbol2', version2, 'Should detect SBOL2 document')
+
+        # Test SBOL3 document detection
+        sbol3_file_xml = os.path.join(TEST_FILES_DIR, 'sbol3_collection.xml')
+        graph3_xml = sbol_utilities.sbol_diff._load_rdf(sbol3_file_xml)
+        version3 = sbol_utilities.sbol_diff._detect_sbol_version(graph3_xml)
+        self.assertEqual('sbol3', version3, 'Should detect SBOL3 document')
+
+        sbol3_file_nt = os.path.join(TEST_FILES_DIR, 'sbol3_collection.nt')
```
jakebeal, line 93: Add a test for a document with mixed SBOL2 and SBOL3 namespaces. You can make this by hand-editing a copy of `sbol3_collection.nt` to be an invalid mixed file.

test/test_sbol_diff.py
```diff
@@ -51,6 +63,129 @@ def test_doc_diff(self):
...
+
+        self.assertIn('different versions', str(context.exception).lower())
+
+    def test_validate_backport_properties_sbol2(self):
+        """Test backport property validation for SBOL2 documents"""
+        clean_file = os.path.join(TEST_FILES_DIR, 'test_attachment_sbol2.xml')
+        clean_graph = sbol_utilities.sbol_diff._load_rdf(clean_file)
+
+        result = sbol_utilities.sbol_diff.validate_backport_properties(clean_graph, 'sbol2')
+        self.assertTrue(result, 'Clean SBOL2 document should pass validation')
```
jakebeal: Also test an invalid document.

test/test_sbol_diff.py
```diff
@@ -51,6 +63,129 @@ def test_doc_diff(self):
...
+        graph3_xml = sbol_utilities.sbol_diff._load_rdf(sbol3_file_xml)
+        version3 = sbol_utilities.sbol_diff._detect_sbol_version(graph3_xml)
+        self.assertEqual('sbol3', version3, 'Should detect SBOL3 document')
+
+        sbol3_file_nt = os.path.join(TEST_FILES_DIR, 'sbol3_collection.nt')
+        graph3_nt = sbol_utilities.sbol_diff._load_rdf(sbol3_file_nt)
+        version3_nt = sbol_utilities.sbol_diff._detect_sbol_version(graph3_nt)
+        self.assertEqual('sbol3', version3_nt, 'Should detect SBOL3 document')
+
+    def test_cross_version_comparison_raises_error(self):
```
jakebeal: Per my note in `sbol_diff.py`, I think this should just be "false" rather than "error"

jakebeal (changes requested):
This is very nice work - just a few minor tweaks needed to finish, I think.

GeneCodeSavvy:
I have made the changes as per your comments @jakebeal

sbol_utilities/sbol_diff.py
```diff
@@ -18,35 +19,32 @@ def _load_rdf(fpath: Union[str, bytes, os.PathLike]) -> rdflib.Graph:
     return graph1
 
 
-def _detect_sbol_version(graph: rdflib.Graph) -> Optional[str]:
+def _detect_sbol_version(graph: rdflib.Graph):
```
jakebeal:
```suggestion
def _detect_sbol_version(graph: rdflib.Graph) -> Optional[str]:
```
Why delete the type information?

sbol_utilities/sbol_diff.py
```diff
@@ -18,35 +19,32 @@ def _load_rdf(fpath: Union[str, bytes, os.PathLike]) -> rdflib.Graph:
...
-            return 'sbol3'
-
-    return None
+    has_v2 = any(sbol2_namespace in ns for ns in namespaces)
+    has_v3 = any(sbol3_namespace in ns for ns in namespaces)
 
+    if has_v2 and has_v3:
+        return None
+    elif has_v2:
+        return 'sbol2'
```
jakebeal: Consider changing these "magic value" constants (both here and elsewhere in the code) into values from a string enum type with a name like SBOLVersion. Using values like "SBOLVersion.SBOL2" instead of raw string values will allow type-checking to be applied to the values and ensure that they are used consistently across different functions.

sbol_utilities/sbol_diff.py
```diff
@@ -112,18 +108,25 @@ def _remove_selective_backport_properties(graph: rdflib.Graph) -> rdflib.Graph:
...
+    :param g1: The first RDF graph to compare.
+    :param g2: The second RDF graph to compare.
+    :param strip_backport_properties: If True, remove backport properties before comparison.
+                                      Defaults to False.
+    :return: A tuple of three graphs: (both, in_g1, in_g2).
+             - both: Statements that are common to both graphs.
+             - in_g1: Statements that are only in the first graph.
+             - in_g2: Statements that are only in the second graph.
+    """
     g1_sbol_version = _detect_sbol_version(g1)
```
jakebeal: These version values aren't used any more and should be deleted.

jakebeal (approved):
Nice work!

GeneCodeSavvy:
@jakebeal  please review

jakebeal (approved):

sbol_utilities/sbol_diff.py
blaze27444: Outdated

sbol_utilities/sbol_diff.py
blaze27444: Outdated

sbol_utilities/sbol_diff.py
blaze27444: Outdated

sbol_utilities/sbol_diff.py
blaze27444: Outdated

sbol_utilities/sbol_diff.py
blaze27444: Outdated

sbol_utilities/sbol_diff.py
blaze27444: Outdated

test/test_sbol_diff.py
blaze27444: Outdated

test/test_sbol_diff.py
blaze27444: Outdated

sbol_utilities/sbol_diff.py
blaze27444: Outdated

sbol_utilities/sbol_diff.py
blaze27444: Outdated

sbol_utilities/sbol_diff.py
blaze27444: Outdated

Issue #364, opened 2025-10-20
The new converter is not accessible through the command line


Gonza10V:
In order to create a test script to compare them we need this feature accesible through the command line.

jakebeal:
fixed with #365

PR #365, opened 2025-10-20
Add ability to invoke new SBOL2/SBOL3 converter from the command line

jakebeal:
The new SBOL2/SBOL converter was not able to be invoked from the command line because the argument for using the new converter was being ignored. This adds its value to the calls to the converter.


Gonza10V (approved):
It worked on my end, both converter worked and the output is slightly different. 57 lines with the new converter and 62 with the old one.
Screenshot 2025-10-20 at 1 52 57 PM

Gonza10V:
This Resolves #364 
@jakebeal you can merge when ready

